PR #4, opened 2023-09-18
Initial repo setup

Siegrift:
Initial repository setup with current implementation of [signed-api](https://github.com/nodaryio/signed-api) and [signed-data-pusher](https://github.com/nodaryio/signed-data-pusher). All of the changes are setup related. Review commit by commit for easier understanding of the actions.

Changes:
- [x] Use pnpm workspaces for pusher and api. I have moved away from initial suggested naming and used `api` and `data-pusher` as workspace names.
- [x] Moved both packages from nodary into this repo
- [x] Improved config files and made them consistent
- [x] Update READMEs

PR #5, opened 2023-09-18
Add husky hooks, CI tests and purge dependencies

Siegrift:
- [x] Setup Husky with pre-push hook - I've only added pre-push hook (because pre-commit is unusable if you do rebase) and I've only included prettier and eslint. Let me know if you prefer also `tsc` and `tests`.
- [x] Add GH actions lint/test
- [x] Make lint and tests pass
- [x] Purge and make consistent use of dependencies. Each workspace defines it's own dependencies, but I moved common dev dependencies to monorepo root. This makes it convenient because the the project doesn't need to install it and all projects will use the same version by default.


packages/api/.eslintignore
```diff
@@ -0,0 +1 @@
+../../.eslintignore
```
andreogle, line 1: Are these files needed?

andreogle (approved):

packages/api/.eslintignore
Siegrift: https://github.com/api3dao/signed-api/pull/4#discussion_r1328618447

PR #7, opened 2023-09-18
Remove serverless and dockerize api

Siegrift:
Removes the Dynamo DB dependency from API and introduces in-memory storage for now. The storage requirements are fairly minimal so we can later plug-in more sophisticated storage if need be. For now, the signed data are kept in a dictionary by Airnode ID and template ID.

The API is now dockerized. There is a simple docker-compose in preparation of adding proper storage (such as Redis or some SQL solution).

With this PR, there should not be any remnants of AWS in API. I've also changed all of the scripts and README. I have **not** tested deploying the Docker on cloud provider, but I tested locally.


.dockerignore
```diff
@@ -0,0 +1,18 @@
+# NOTE: Keep in sync with .gitignore.
+#
+# From: https://shisho.dev/blog/posts/how-to-use-dockerignore/
+#
+# In .gitignore, the file or directory name is ignored in any hierarchy below the .gitignore file, but in .dockerignore,
```
Siegrift, line 5: @andreogle This is actually my first time doing Docker stuff and I am not a bit sad that every tool has a different resolution for their ignore files... While with prettier and ESlint the symlink solution is "usable", you can't make it work for Docker ignore.

packages/api/.env.example
```diff
@@ -1,2 +1,2 @@
-HTTP_API_ID=
-MAX_BATCH_SIZE=
\ No newline at end of file
+MAX_BATCH_SIZE=10
+PORT=8090
```
Siegrift, line 2: The PORT was missing in README and examle.env and I was lazy to create a separate commit/pr for it.

packages/api/README.md
```diff
@@ -4,29 +4,11 @@ A service for storing and accessing signed data. It provides endpoints to handle
...
-```bash
-pnpm run removeDeployment
-```
-
-## Public Endpoint
-
-The API is publicly accessible at the following endpoint:
-
-- https://pool.nodary.io
+TODO: Write example how to deploy on AWS (and maybe other cloud providers as well).
```
Siegrift, line 11: I don't necessarily plan to resolve this in this PR. Wdyt?

packages/api/package.json
```diff
@@ -7,12 +7,10 @@
...
   "main": "index.js",
   "scripts": {
-    "deploy": "serverless deploy",
-    "removeDeployment": "serverless remove",
-    "dynamodb-local-up": "docker compose -f docker-compose.dynamodb-local.yml up -d ",
-    "dynamodb-local-down": "docker compose -f docker-compose.dynamodb-local.yml down ",
-    "init-tables": "LOCAL_DEV=true npx ts-node scripts/init-tables.ts",
-    "start-local": "LOCAL_DEV=true npx ts-node src/local-server.ts",
+    "dev": "pnpm ts-node src/local-server.ts",
+    "docker:start": "docker compose --file docker/docker-compose.yml up",
```
Siegrift, line 11: Since we don't have any volumes, these seemed enough. Did I miss some useful command?

andreogle: I think this is good 👍🏻

packages/api/src/in-memory-cache.ts
andreogle, line 1: Nice work 👍🏻

packages/api/README.md
andreogle: Do you mean add instructions for deployment in this PR? No, I don't think that should be done as part of this PR

packages/api/src/local-server.ts
```diff
@@ -1,39 +1,44 @@
...
-dotenv.config();
 import express from 'express';
-import { APIGatewayProxyEvent } from 'aws-lambda';
 import { getData, listAirnodeAddresses, batchUpsertData, upsertData } from './handlers';
 
+dotenv.config();
 const { PORT } = process.env;
 
-const port = PORT || 8090;
+const port = PORT;
```
andreogle, line 8: I think port should have a default value if it's not set as an env variable

```suggestion
const port = process.env.PORT || 8090;
```

andreogle (approved):

packages/api/README.md
Siegrift: Yeah, that's what I meant.

packages/api/docker/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Step 1: Build the TypeScript application
+FROM node:18-alpine AS builder
+
+WORKDIR /usr/src/app
+COPY pnpm-lock.yaml package.json .
```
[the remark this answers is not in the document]
Siegrift: No 👍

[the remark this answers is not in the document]
Siegrift: Actually I understand why I specified it that way. I got inspired by a similar Dockerfile and it had this copying in two steps. The reason is that docker caches the steps. So if you first copy lock file and package.json and then run `pnpm install` this will only run if the files change. When you copy everything, the install step will re-run even if you change some .ts file... I'd say we don't care because the dockerized version will be only used for deployment and not development.

packages/api/src/local-server.ts
Siegrift: I put the default value inside the `.env.example` and the README tells you to copy it. Or am I missing something?

andreogle: I guess I just typically leave it out and just use whatever the default is set to. Fine with leaving it as required too

PR #24, opened 2023-09-20
Refactor pusher

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/15

Review commit by commit. I separated the actions I was doing into separate commits.

## Images

The first command builds the docker, looking for config files inside `data-pusher/config` directory. The second command starts the image and correctly errors out on incorrect configuration error.

image

PR #26, opened 2023-09-20
Small API improvements

Siegrift:
Some improvements I noticed during work on https://github.com/api3dao/signed-api/pull/24

- Enables source map support
- Creates a proper `index.ts` entrypoint that starts the server.


packages/api/src/server.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  app.put('/', async (req, res) => {
+    const result = await upsertData({
+      body: JSON.stringify(req.body),
+      queryParams: {},
+    });
+    res.status(result.statusCode).header(result.headers).send(result.body);
+  });
+
+  app.post('/', async (req, res) => {
```
andreogle, line 22: Using `post`  for batching upserts (while `put` is used for single upserts on the same path) doesn't seems so intuitive to me. I think these should be different paths using the same HTTP method

andreogle (approved):

PR #27, opened 2023-09-20
Update deps

Siegrift:
Updates all of the dependencies in 1 go instead of merging them 1 by 1 using renovate.

Also, we do not want to migrate to new `ethers` major, because that is basically a new library, so I've modified renovate config to ignore it.


andreogle:
Would you mind merging `main` into this branch? I can review once that's done

andreogle (approved):

Issue #21, opened 2023-09-20
Allow API to serve delayed data


Siegrift:
See: https://api3workspace.slack.com/archives/C05S589E7B4/p1695136091683809

What this means is that the API should be (statically = at startup time) configurable to serve multiple versions of data. API will need to keep older signed data. However, it should automatically prune data that is not needed.

This entails:
- [x] Modify API configuration to allow serving multiple data types (delayed, non-delayed).
- [x] Remove upserting HTTP endpoint
- [x] Keep all beacon signed data to allow serving the data point
- [x] Document the configuration and how to serve multiple data types.

bbenligiray:
To be specific, if a 15 seconds delay is enforced and the signed API has 14 and 16 seconds-old data, it returns the 16 seconds-old data.

PR #30, opened 2023-09-20
Run signed API together with pusher

Siegrift:
As a preparation for https://github.com/api3dao/signed-api/issues/21 I wanted to run the services together, but it took me some configuration effort so I decided to make a PR out of my work to save others a bit more time.

This PR:
- Changes the pusher example configuration to make it easy to run the services together.
- Documents how to run them together and basic dev instructions.
- Adds very basic logging and leaves some TODOs for later

### DIFF
--- a/packages/api/src/server.ts
+++ b/packages/api/src/server.ts
@@ -12,6 +12,9 @@ export const startServer = () => {
   app.use(express.json());
 
   app.put('/', async (req, res) => {
+    // eslint-disable-next-line no-console
+    console.log('Received request "PUT /"', req.body, req.params, req.query);
+
     const result = await upsertData({
       body: JSON.stringify(req.body),
       queryParams: {},
@@ -20,6 +23,9 @@ export const startServer = () => {
   });
 
   app.post('/', async (req, res) => {
+    // eslint-disable-next-line no-console
+    console.log('Received request "POST /"', req.body, req.params, req.query);
+
     const result = await batchUpsertData({
       body: JSON.stringify(req.body),
       queryParams: {},
@@ -28,6 +34,9 @@ export const startServer = () => {
   });
 
   app.get('/:airnode', async (req, res) => {
+    // eslint-disable-next-line no-console
+    console.log('Received request "GET /:airnode"', req.body, req.params, req.query);
+
     const result = await getData({
       body: '',
       queryParams: { airnode: req.params.airnode },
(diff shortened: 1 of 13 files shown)


packages/api/src/server.ts
```diff
@@ -12,6 +12,9 @@ export const startServer = () => {
   app.use(express.json());
 
   app.put('/', async (req, res) => {
+    // eslint-disable-next-line no-console
+    console.log('Received request "PUT /"', req.body, req.params, req.query);
```
andreogle, line 16: I think we should rather use an established logger like [morgan](https://github.com/expressjs/morgan) (doesn't seem to be maintained much, but neither is express really) or [winston](https://github.com/winstonjs/winston). It's been a long time since I've configured express so maybe there are newer options out there

packages/data-pusher/src/make-request.ts
```diff
@@ -267,17 +267,23 @@ export const postSignedApiData = async (group: SignedApiNameUpdateDelayGroup) =>
...
-    logger.warn(`Failed to post payload to update signed API. Err: ${goRes.error}`, logOptions);
+    logger.warn(
+      // See: https://axios-http.com/docs/handling_errors
+      `Failed to post payload to update signed API. Err: ${goRes.error}, axios response: ${goRes.error.response}`,
+      logOptions
+    );
     return;
   }
-  logger.info(`Signed datas (${goRes.data.data.count.toString()}) were pushed to pool.`, logOptions);
+  logger.info(`Signed datas (${goRes.data.count.toString()}) were pushed to pool.`, logOptions);
```
andreogle, line 288:
```suggestion
  logger.info(`Pushed ${goRes.data.count.toString()} signed data updates to the pool.`, logOptions);
```

packages/api/src/server.ts
andreogle: Ok sounds good. It's wild to me how much configuration winston needs 😄

andreogle (approved):

packages/api/src/server.ts
Siegrift: You don't need a lot of it to setup a basic logger, but I spent some time tweaking the options to make it work nicely (and also our logging style).

Siegrift: Created https://github.com/api3dao/signed-api/issues/32

Issue #6, opened 2023-09-18
Consider more strict TS settings


Siegrift:
https://github.com/api3dao/signed-api/pull/4/files#r1328594362 and https://github.com/api3dao/signed-api/pull/4/files#r1328595642

PR #31, opened 2023-09-21
Enable stricter TS checks, fix validation issues

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/6

Also resolves the TODOs created in https://github.com/api3dao/signed-api/pull/30

### DIFF
--- a/packages/api/src/in-memory-cache.ts
+++ b/packages/api/src/in-memory-cache.ts
@@ -5,7 +5,7 @@ const signedDataCache: Record<string /* Airnode ID */, Record<string /* Template
 // The API is deliberately asynchronous to mimic a database call.
 export const getBy = async (airnodeId: string, templateId: string) => {
   if (!signedDataCache[airnodeId]) return null;
-  return signedDataCache[airnodeId][templateId] ?? null;
+  return signedDataCache[airnodeId]![templateId] ?? null;
 };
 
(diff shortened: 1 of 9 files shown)


packages/api/src/in-memory-cache.ts
```diff
@@ -5,7 +5,7 @@ const signedDataCache: Record<string /* Airnode ID */, Record<string /* Template
 // The API is deliberately asynchronous to mimic a database call.
 export const getBy = async (airnodeId: string, templateId: string) => {
   if (!signedDataCache[airnodeId]) return null;
-  return signedDataCache[airnodeId][templateId] ?? null;
+  return signedDataCache[airnodeId]![templateId] ?? null;
```
andreogle, line 8: I would have assumed this wouldn't be necessary since `!signedDataCache[airnodeId]` was just checked in the line above 🤔

andreogle (approved):

packages/api/src/in-memory-cache.ts
Siegrift: You can have getters which may not return the same thing so TS will be extra cautios. THis would work:
```
  const x = signedDataCache[airnodeId];
  if (!x) return null;
  return x[templateId] ?? null;
```

but I don't wanna do that.

Issue #15, opened 2023-09-19
Refactor pusher


Siegrift:
See: https://api3workspace.slack.com/archives/C05S589E7B4/p1695117126655149?thread_ts=1695115119.548779&cid=C05S589E7B4

- [x] Remove serverless
- [x] Remove referencing Airseeker
- [x] Import from Airnode whenever we can
- [x] Dockerize
- [x] Document (at least dockerization, the features documentation will be added later)

PR #33, opened 2023-09-21
Refactor logic for making requests

Siegrift:
As part of https://github.com/api3dao/signed-api/issues/15 I wanted to import whatever we can from Airnode or at least move it to a dedicated directory. I noticed now, that I missed some Airnode logic in the `make-requests.ts` file. Unfortunately, processing logic is not exported from Airnode so we have to leave it copy pasted for now.

This PR:
- Uses `processSuccessfulApiCall` from Airnode.
- Moves processing logic to unexported Airnode features folder - note that the original Airnode implementation slightly changed in the meantime.
- The pusher related logic for making requests has been split into two files
  - `data-provider.ts` - for making template requests against data provider API.
  - `signed-api.ts` - for signing the template requests and making requests to the signed API.


andreogle (approved):
👍🏻

PR #34, opened 2023-09-24
Allow API to serve delayed data

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/21

## Rationale

The main change is how the signed data is stored. We no longer store the latest data for each beacon, but we keep an array of data instead. This makes it easy to compute the freshest data for each endpoint while keeps the memory quite generic. With this change it is no longer possible to update signed data - pusher can only append signed data to this service which **always accepts** valid data (and may decide to "forget" about it is not needed). 

To support multiple endpoints, it is clear we need some configuration. I opted for `signed-api.json` with basic schema that allows to configure an endpoint with a delay (currently without auth). The signed API ignores all data that is too fresh and was submitted after `"Date.now() - delayInMs"`. I then realized, we no longer need the ENV file and can instead merge this with the config (so that we have a single file to maintain).

I added some basic tests and minor refactors to simplify the code.

Note, that I have not added support for pruning old data and instead created a separate issue https://github.com/api3dao/signed-api/issues/35 since the PR is quite big already.


packages/api/config/signed-api.example.json
andreogle, line 1: 👍🏻

andreogle (approved):

PR #38, opened 2023-09-25
Add hot reload

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/29

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "build": "tsc --project .",
     "clean": "rm -rf coverage dist",
-    "dev": "nodemon --ext ts,js,json,env  --exec \"pnpm run start\"",
+    "dev": "nodemon --ext ts,js,json,env  --exec \"pnpm ts-node src/index.ts\"",
     "docker:build": "docker compose --file docker/docker-compose.yml build",
     "docker:start:detach": "docker compose --file docker/docker-compose.yml up --detach",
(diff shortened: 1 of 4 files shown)


packages/api/package.json
```diff
@@ -18,13 +18,16 @@
     "eslint:fix": "eslint . --ext .js,.ts --fix",
     "prettier:check": "prettier --check \"./**/*.{js,ts,md,json}\"",
     "prettier:fix": "prettier --write \"./**/*.{js,ts,md,json}\"",
+    "start": "pnpm ts-node src/index.ts",
```
andreogle, line 21: Is it not a bit more intuitive/common to have just `dev` and `start` scripts? The `dev` command could have everything inline: `"dev": nodemon --ext ts,js,json,env --exec \"ts-node src/index.ts\"

It's not obvious to me when someone would want to use `pnpm start` over `pnpm dev`

packages/api/package.json
```diff
@@ -18,13 +18,16 @@
...
+    "start:prod": "pnpm node dist/src/index.js",
```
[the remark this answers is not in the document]
Siegrift: No, we don't.

packages/api/package.json
Siegrift: Nodemon will restart the server everytime a file changes in the current folder, so I though maybe there could be and use case (e.g. I do not want to reload while working on tests). That said, I don't mind removing the `start` script completely.

Issue #25, opened 2023-09-20
Plan and implement unit tests


Siegrift:
The API package has reasonable unit test coverage and does not need additional tests.

The pusher package would benefit from testing the logic for making requests, that is tests for these files:
- [x] https://github.com/api3dao/signed-api/tree/83b6b1624399289d4ae71864b6e2071372272fec/packages/data-pusher/src/api-requests

PR #44, opened 2023-09-26
Add unit tests for api-requests modules

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/25

- During the test implementation I noticed that some logs that were a bit misleading so I improved them.
- There were some unnecessary types and wrong typings which I also corrected.
- Since I made the above changes, pay a bit more attention to production files (not ending with .test.ts and not inside test folder)
- There are some other changes, explained inside the PR.


packages/data-pusher/src/api-requests/data-provider.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  nodaryTemplateRequestErrorResponse,
+  nodaryTemplateRequestResponseData,
+  nodaryTemplateResponses,
+} from '../../test/fixtures';
+
+describe(makeTemplateRequests.name, () => {
+  it('makes a single template request for multiple beacons', async () => {
+    const state = stateModule.getInitialState(config);
+    jest.spyOn(stateModule, 'getState').mockReturnValue(state);
+    const logger = createMockedLogger();
```
Siegrift, line 17: Tip: If you want to know what is actually logged, you can do:
```ts
(logger as any).debug.mockImplementation(console.log)
```

which will forward the logs to console log. I used that a bit.

packages/data-pusher/src/api-requests/data-provider.ts
```diff
@@ -1,31 +1,30 @@
...
 import { SignedApiUpdate, TemplateId } from '../validation/schema';
 import { getLogger } from '../logger';
 
-type TemplateResponse = [TemplateId, node.HttpGatewayApiCallSuccessResponse];
-type TemplateResponses = TemplateResponse[];
+export type TemplateResponse = [TemplateId, node.HttpGatewayApiCallSuccessResponse];
 
 export const callApi = async (payload: node.ApiCallPayload) => {
-  getLogger().debug('Preprocessing API call payload', { aggregateApiCall: payload.aggregatedApiCall });
+  getLogger().debug('Preprocessing API call payload', pick(payload.aggregatedApiCall, ['endpointName', 'oisTitle']));
```
Siegrift, line 12: The example file is uses pre/post processing to only make a single API call and if you log the full `aggregateApiCall` you also log `parameters` (of a single beacon) which is confusing, so I only log endpointName and oisTitle.

packages/data-pusher/src/api-requests/data-provider.ts
```diff
@@ -1,31 +1,30 @@
...
   const processedPayload = await preProcessApiSpecifications(payload);
-  getLogger().debug('Performing API call', { aggregateApiCall: payload.aggregatedApiCall });
+  getLogger().debug('Performing API call', { processedPayload: processedPayload });
```
Siegrift, line 14: This was a bug before - we should have logged the processed payloed.

packages/data-pusher/src/api-requests/signed-api.ts
```diff
@@ -46,19 +51,31 @@ export const postSignedApiData = async (group: SignedApiNameUpdateDelayGroup) =>
...
-      { ...logContext, axiosResponse: goRes.error.response }
+      { ...logContext, axiosResponse: goAxiosRequest.error.response }
     );
-    return;
+    return { success: false };
   }
-  getLogger().info(`Pushed signed data updates to the pool.`, { ...logContext, count: goRes.data.count });
+
+  getLogger().debug('Parsing response from the data registry.', { ...logContext, axiosResponse: goAxiosRequest.data });
+  const parsedResponse = dataRegistryResponseSchema.safeParse(goAxiosRequest.data);
```
Siegrift, line 64: This was not handled correctly previously.

packages/data-pusher/src/index.ts
```diff
@@ -9,15 +9,13 @@ import { loadConfig } from './validation/config';
 import { initiateFetchingBeaconData } from './fetch-beacon-data';
 import { initiateUpdatingSignedApi } from './update-signed-api';
 import { initializeState } from './state';
-import { initializeWallet } from './wallets';
```
Siegrift, line 12: This broke the tests, since the `getInitialState` function essentially returned invalid state (wallet was missing).

packages/data-pusher/test/fixtures.ts
Siegrift, line 1: I moved all of the fixtures to this file. It makes the tests a bit cleaner and we might use these in case we want to make an e2e test later on.

Siegrift:
@andreogle I anticipated we will do https://github.com/api3dao/signed-api/issues/45 and renamed some of the variables to data registry for clarity. I need to revert this back to the original naming (that's why I added the WIP).

andreogle (approved):

PR #51, opened 2023-09-27
Dynamically load api config and refactor the logger

andreogle:
This branch originally started out for getting the Cloudformation file setup, but I ended up using it for refactoring the config files based on the discussions. This work can continue separately.

### Changes

1. Removes `logger` from both config files in favour of env variables
2. The `api` package will now dynamically fetch and cache the logger based on `CONFIG_SOURCE`. The data-pusher will follow the same pattern later.
3. Makes the logger usage consistent in across both packages

Let me know what you think


packages/common/src/logger/index.ts
```diff
@@ -2,44 +2,43 @@ import winston from 'winston';
...
 export const logLevelSchema = z.union([z.literal('debug'), z.literal('info'), z.literal('warn'), z.literal('error')]);
 
 export type LogLevel = z.infer<typeof logLevelSchema>;
 
 export const logConfigSchema = z.object({
-  type: logTypeSchema,
-  styling: logStylingSchema,
+  colorize: z.boolean(),
+  enabled: z.boolean(),
+  format: logFormatSchema,
```
andreogle, line 16: I changed these names to make them a bit more intuitive (imo) and follow the underlying `winston` naming.

packages/data-pusher/src/api-requests/signed-api.test.ts
```diff
@@ -62,7 +58,7 @@ describe(postSignedApiData.name, () => {
     );
     jest.spyOn(stateModule, 'getState').mockReturnValue(state);
     const logger = createMockedLogger();
-    jest.spyOn(loggerModule, 'getLogger').mockReturnValue(logger);
+    jest.spyOn(loggerModule, 'logger').mockReturnValue(logger);
```
andreogle, line 61: I'm struggling to figure out how to mock the exported constant (not a function anymore). Any suggestions? Mocking always seems to a massive pain with Jest 😢

packages/common/src/logger/index.ts
Siegrift: I don't mind, but change the documentation as well.

packages/data-pusher/src/api-requests/signed-api.test.ts
Siegrift: Yeah. You can't really mock exported objects without mocking the full module, which I dislike.

Siegrift (commented):
I actually quite liked that the logger configuration was specified in the config file, because everything was configured in a single place.

I see the issue of loading file before initializing logger, but it is not a big deal considering we want to exit the app when config loading fails.

packages/data-pusher/src/api-requests/signed-api.test.ts
Siegrift: What you can do is mock the `logger.debug` since that is a function.

Siegrift:
@andreogle I've made two notable changes.

1. Fix TS configuration by removing `composite` which resulted in outdated TS cache. The composite build is setup on Airnode, which is much larger and we don't need such setup here.
2. Improve validation around the ENVs. We want to fail asap when the user supplies invalid configuration. This is done by flowing the ENVs through zod schema. This way, we also get the ENVs in a more typesafe way (and can apply defaults). Using the `loadEnv` should be used everytime you need some ENV from process.

PR #68, opened 2023-10-05
WIP: Migrate to common ESlint configuration

Siegrift:
Related to https://github.com/api3dao/commons/pull/3

## What to review

I am opening the PR as a draft, that will fail tests. I have used verdaccio to publish the [commons](https://github.com/api3dao/commons) locally and use the ESlint configuration.

The point is to see what the code after migration looks like. Note, that many changes were done automatically via `eslint --fix` so do not take it as an estimate of how difficult the migration is. Notice, that the repo uses ALL rules from the common configuration.


packages/api/src/handlers.test.ts
```diff
@@ -1,29 +1,33 @@
...
 import * as cacheModule from './cache';
 import * as configModule from './config';
 import { batchInsertData, getData, listAirnodeAddresses } from './handlers';
-import { createSignedData, generateRandomWallet } from '../test/utils';
-
-afterEach(() => {
-  cacheModule.setCache({});
-});
 
+// eslint-disable-next-line jest/no-hooks
```
andreogle, line 12: Not sure how I feel about this rule. `beforeEach` and the similar hooks can be pretty useful.

Although, the "correct" version does also seem pretty decent: https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-hooks.md

packages/api/src/handlers.ts
```diff
@@ -1,76 +1,85 @@
...
+  }
 
   // Ensure there is at least one signed data to push
   const batchSignedData = goValidateSchema.data;
   if (isEmpty(batchSignedData)) return generateErrorResponse(400, 'No signed data to push');
 
   // Check whether the size of batch exceeds a maximum batch size
   const { maxBatchSize, endpoints } = getConfig();
-  if (size(batchSignedData) > maxBatchSize)
+  if (size(batchSignedData) > maxBatchSize) {
```
andreogle, line 33: 👍🏻 Multiline blocks without curly braces bugged me 😬

packages/api/src/schema.test.ts
```diff
@@ -96,7 +98,7 @@ describe('env config schema', () => {
     expect(() => envConfigSchema.parse(env)).not.toThrow();
   });
 
-  it('AWS_REGION is set when CONFIG_SOURCE is aws-s3', () => {
+  it('aWS_REGION is set when CONFIG_SOURCE is aws-s3', () => {
```
andreogle, line 101: I think this isn't correct 😅

andreogle (commented):
I really like a lot of these changes. Nice work

packages/api/src/schema.test.ts
Siegrift: Yeah, I noticed it too late (and fixed in the logger PR).

There is and ESLint rule for jest that mandages lowercase starting character for `it`. I am not too keen on keeping it.

PR #69, opened 2023-10-06
Use common logger

Siegrift:
Because commons only had one functionality (the logger) it is now removed in favor of the logger from commons.


packages/api/src/schema.test.ts
```diff
@@ -98,7 +98,7 @@ describe('env config schema', () => {
     expect(() => envConfigSchema.parse(env)).not.toThrow();
   });
 
-  it('aWS_REGION is set when CONFIG_SOURCE is aws-s3', () => {
+  it('the AWS_REGION is set when CONFIG_SOURCE is aws-s3', () => {
```
andreogle, line 101: 👍🏻

andreogle (approved):

PR #81, opened 2023-10-11
Update @api3/commons

Siegrift:
Relates to https://github.com/api3dao/commons/issues/14 and https://github.com/api3dao/commons/pull/15


packages/api/src/schema.ts
```diff
@@ -59,8 +59,36 @@ export const envBooleanSchema = z.union([z.literal('true'), z.literal('false')])
 export const envConfigSchema = z
   .object({
     LOG_COLORIZE: envBooleanSchema.default('false'),
-    LOG_FORMAT: logFormatSchema.default('json'),
-    LOG_LEVEL: logLevelSchema.default('info'),
+    LOG_FORMAT: z
+      .string()
+      .transform((value, ctx) => {
```
Siegrift, line 64: @andreogle actually, we do not need to validate this because the logger does. So without this the implementation will be more straightforward.

PR #94, opened 2023-10-20
Implement node version

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/82

## Rationale

Added `nodeSettings.nodeVersion` similar to the Airnode. I am not entirely happy with the naming, but sticked to the on used in Airnode. Also, I put the node version inside `nodeSettings` in anticipation for other fields (e.g. stage).

I also noticed that I didn't completely remove source maps in https://github.com/api3dao/signed-api/pull/64 so I removed the (no longer necessary) `--enable-source-maps` flag.

### DIFF
--- a/packages/e2e/src/pusher/pusher.json
+++ b/packages/e2e/src/pusher/pusher.json
@@ -107,5 +107,8 @@
       ]
     }
   ],
-  "apiCredentials": []
+  "apiCredentials": [],
+  "nodeSettings": {
+    "nodeVersion": "0.1.0"
+  }
 }
--- a/packages/e2e/src/pusher/pusher.json
+++ b/packages/e2e/src/pusher/pusher.json
@@ -1,6 +1,4 @@
 {
-  "airnodeWalletMnemonic": "diamond result history offer forest diagram crop armed stumble orchard stage glance",
-  "rateLimiting": { "Mock API": { "maxConcurrency": 25, "minTime": 0 } },
   "templates": {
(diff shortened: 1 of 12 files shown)


bbenligiray:
In the Airnode config `airnodeWalletMnemonic` is also under `nodeSettings` https://github.com/api3dao/airnode/blob/master/packages/airnode-node/config/config.example.json#L68
`rateLimiting` only exists in Airseeker and it's at the config root for the lack of a `nodeSettings` field (https://github.com/api3dao/airseeker/blob/main/config/airseeker.example.json#L3), but I'd argue that if there is a `nodeSettings` field, `rateLimiting` should go under it

packages/e2e/src/pusher/pusher.json
```diff
@@ -107,5 +105,10 @@
       ]
     }
   ],
-  "apiCredentials": []
+  "apiCredentials": [],
+  "nodeSettings": {
+    "nodeVersion": "0.1.0",
+    "airnodeWalletMnemonic": "diamond result history offer forest diagram crop armed stumble orchard stage glance",
```
andreogle, line 111: I hope this isn't linked to any real funds (I see it was already present anyway 😅)

andreogle (approved):

packages/e2e/src/pusher/pusher.json
Siegrift: Feel free to send any 💵 :D

Issue #95, opened 2023-10-23
Add stage and pusher version fields to the pusher config


metobom:
We need to identify pusher deployments and we need a field named stage ~and a version of the app as it is in the Airnode config.~

Siegrift:
We already added version in https://github.com/api3dao/signed-api/issues/82. Or do you mean something different?

(The stage is not added yet, though)

metobom:
No, it is the same thing, thanks.

Issue #96, opened 2023-10-23
Add logs to use as heartbeat


metobom:
Pusher needs to log the below with a minute frequency: 

```
pusherConfigHash = hash(config) // config is not interpolated
heartbeatSignature = sign(
  hash(
    airnodeAddress,
    stage, 
    pusherVersion,
    currentTimestamp,
    deploymentTimestamp,
    pusherConfigHash
  )
) // deploymentTimestamp: timestamp that pusher started to work
log = `${airnodeAddress} - ${stage} - ${pusherVersion} - ${currentTimestamp} - ${deploymentTimestamp} - ${pusherConfigHash} - ${heartbeatSignature}` 
```
This is required because we need to make sure
* the API provider deployed the correct thing
* pusher is alive

and also we need to keep track of deployments in [api-integrations](https://github.com/api3dao/api-integrations).

This is blocked by https://github.com/api3dao/signed-api/issues/95 because it requires `stage` and `pusherVersion`.

Siegrift:
For `deploymentTimestamp` is enough to take a timestamp when the pusher is started ([here](https://github.com/api3dao/signed-api/blob/fb5be4ebf8b6c405a948cc58b697b3a115e50111/packages/pusher/src/index.ts#L6C1-L6C1))?

metobom:
Yes, it is enough.

PR #99, opened 2023-10-29
Add stage to signed API

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/95

The purpose of the stage field is to send it with logs as a "heartbeat" in issue https://github.com/api3dao/signed-api/issues/96

### DIFF
--- a/packages/pusher/src/validation/schema.ts
+++ b/packages/pusher/src/validation/schema.ts
@@ -253,7 +253,7 @@ export const nodeSettingsSchema = z.object({
   rateLimiting: rateLimitingSchema,
   stage: z
     .string()
-    .regex(/^[\da-z-]{1,16}$/, 'Only lowercase letters, numbers and hyphens are allowed (max 16 characters)'),
+    .regex(/^[\da-z-]{1,256}$/, 'Only lowercase letters, numbers and hyphens are allowed (max 256 characters)'),
 });
 
 export type NodeSettings = z.infer<typeof nodeSettingsSchema>;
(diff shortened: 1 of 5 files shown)


packages/pusher/src/validation/schema.ts
```diff
@@ -251,6 +251,9 @@ export const nodeSettingsSchema = z.object({
   nodeVersion: z.string().refine((version) => version === packageJson.version, 'Invalid node version'),
   airnodeWalletMnemonic: z.string().refine((mnemonic) => ethers.utils.isValidMnemonic(mnemonic), 'Invalid mnemonic'),
   rateLimiting: rateLimitingSchema,
+  stage: z
+    .string()
+    .regex(/^[\da-z-]{1,16}$/, 'Only lowercase letters, numbers and hyphens are allowed (max 16 characters)'),
```
Siegrift, line 256: I took inspiration from Airnode, where we use this validation for stage.

Siegrift: cc: @metobom

bbenligiray: Airnode needs a very short `stage` because it's used in cloud provider resource names. There is no such restriction here, and `stage` is our only identifier (while Airnode has `cloudProvider` and `region` in addition) so we'll want it to be a lot longer.

Siegrift: 256 chars?

metobom: 256 should be sufficient.

PR #100, opened 2023-10-29
Add heartbeat log

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/96

### DIFF
--- a/packages/pusher/src/heartbeat/heartbeat.ts
+++ b/packages/pusher/src/heartbeat/heartbeat.ts
@@ -5,6 +5,8 @@ import { logger } from '../logger';
 import { getState } from '../state';
 import { loadRawConfig } from '../validation/config';
 
+import { heartbeatLogger } from './logger';
+
 export const initiateHeartbeat = () => {
   logger.debug('Initiating heartbeat loop');
   setInterval(async () => {
@@ -47,8 +49,7 @@ export const logHeartbeat = async () => {
   const heartbeatSignature = await signHeartbeat(airnodeWallet, heartbeatPayload);
   const heartbeatLog = [...heartbeatPayload, heartbeatSignature].join(' - ');
 
-  // We ensure in config validation that INFO level logs are not disabled. The logs are sent to API3 for validation
-  // (that the data provider deployed deployed the correct configuration) and monitoring purposes (whether the instance
-  // is running).
-  logger.info(heartbeatLog);
+  // The logs are sent to API3 for validation (that the data provider deployed deployed the correct configuration) and
+  // monitoring purposes (whether the instance is running).
+  heartbeatLogger.info(heartbeatLog);
 };
--- a/packages/pusher/src/heartbeat/heartbeat.ts
+++ b/packages/pusher/src/heartbeat/heartbeat.ts
@@ -1,0 +1,54 @@
+import { go } from '@api3/promise-utils';
+import { ethers } from 'ethers';
+
+import { logger } from '../logger';
+import { getState } from '../state';
+import { loadRawConfig } from '../validation/config';
+
+export const initiateHeartbeat = () => {
+  logger.debug('Initiating heartbeat loop');
+  setInterval(async () => {
+    const goLogHeartbeat = await go(logHeartbeat);
+    if (!goLogHeartbeat.success) logger.error('Failed to log heartbeat', goLogHeartbeat.error);
+  }, 1000 * 60); // Frequency is hardcoded to 1 minute.
+};
+
+export const signHeartbeat = async (airnodeWallet: ethers.Wallet, heartbeatPayload: unknown[]) => {
+  logger.debug('Signing heartbeat payload');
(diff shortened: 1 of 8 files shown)


packages/pusher/test/fixtures.ts
```diff
@@ -183,3 +185,26 @@ export const signedApiResponse: Partial<AxiosResponse> = {
   },
   data: { count: 3 },
 };
+
+export const parseHeartbeatLog = (logMessage: string) => {
```
Siegrift, line 189: @metobom You will need to detect the heartbeat logs and then parse it. This would be the way to do that.

packages/pusher/src/heartbeat/heartbeat.ts
```diff
@@ -0,0 +1,55 @@
...
+  const signaturePayload = ethers.utils.arrayify(createHash(JSON.stringify(heartbeatPayload)));
+  return airnodeWallet.signMessage(signaturePayload);
+};
+
+export const createHash = (value: string) => ethers.utils.keccak256(ethers.utils.toUtf8Bytes(value));
+
+export const logHeartbeat = async () => {
+  logger.debug('Creating heartbeat log');
+
+  const rawConfig = loadRawConfig(); // We want to log the raw config, not the one with interpolated secrets.
```
andreogle, line 29: 👍🏻

packages/pusher/src/heartbeat/index.ts
andreogle, line 1: Nit: why not just have `src/heartbeat.ts` (and `src/heartbeat.test.ts`)?

andreogle (approved):

packages/pusher/src/heartbeat/index.ts
Siegrift: I like barrel imports, so that you can `import {xyz} from './src/heartbeat'` instead of `import {xyz} from './src/heartbeat/heartbeat'`

andreogle: You get that same import line with just `src/heartbeat.ts` though?

I'm personally finding myself moving away from barrel files haha. They aren't nice for file searching (lots of `index.ts` files) and the [performance problems they bring when there are lots of them](https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-7/) (obviously not an issue here). Next.js has jaded me with these kinds of things 😬 . It's also kinda weird when there is only other 1 file in the folder

Siegrift: I see, you mean why I have it in separate folder. 

I prefer to use a folder if it encapsulates a feature. It reduces the number of files in parent directory and it's also easier to understand what feature the module belongs to (the folder name should tell you).

packages/pusher/test/fixtures.ts
Siegrift: Btw. one thing not mentioned (but you are probably aware) is that the hash of the config needs to be verified as well. In my implementation I only do that inside the test [here](https://github.com/api3dao/signed-api/blob/639718399d91933e69315fee143b709c64aff22a/packages/pusher/src/heartbeat/heartbeat.test.ts#L58).

packages/pusher/src/heartbeat/heartbeat.ts
```diff
@@ -0,0 +1,55 @@
...
+  const heartbeatPayload = [
+    airnodeWallet.address,
+    stage,
+    nodeVersion,
+    currentTimestamp.toString(),
+    deploymentTimestamp.toString(),
+    rawConfigHash,
+  ];
+  const heartbeatSignature = await signHeartbeat(airnodeWallet, heartbeatPayload);
+  const heartbeatLog = [...heartbeatPayload, heartbeatSignature].join(' - ');
```
bdrhn9, line 50: Doesn't it better to have some words indicates that it is heartbeat log? Like:
```
[Heartbeat] - 0xdAD65588187bC1a76C482174e25ACcED1413d2a1 - dev - 0.1.0 - 1699020775 - 1699020712 - 0xe872866f40b6a397b5bb5fe86b94809ba72e237633200c05d18ed345631fbd3b - 0x52843fc7b68cf2bacc6084929e6d77dbd71541d9b83838d078aaa2d7522eefec75b85a9f7628db787263d2dec93cb6d4194111411ec747d3b91d133c26bcb8341c
```

and log doesn't self descriptive so I'd prefer to have:

```json
{
"airnode": "0xdAD65588187bC1a76C482174e25ACcED1413d2a1",
"stage": "dev",
"nodeVersion": "0.1.0",
...
"configHash": "0x52843fc7b68cf2bacc6084929e6d77dbd71541d9b83838d078aaa2d7522eefec75b85a9f7628db787263d2dec93cb6d4194111411ec747d3b91d133c26bcb8341c"
}
```

Because I'm querying logs and searching this log type among all, it's not easy with this schema.

Siegrift: Yeah, no problem. I think that would be better.

I can make the log message to be "Sending heartbeat" and specify all those fields as log context (which is JSON so it should be queryable). I've created https://github.com/api3dao/signed-api/issues/107

Issue #98, opened 2023-10-24
Add whitelisted Airnode addresses to the signed API configuration


metobom:
Since anyone can push data in the current implementation, it is possible for a signed API to become messy. So adding a new field that includes whitelisted Airnode addresses would solve this problem.

PR #101, opened 2023-10-30
Implement Airnode whitelisting

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/98

## Rationale

Adding the configuration option for whitelisting addresses is simple, but there is a special case - how to whitelist all? We can have a special meaning for empty array, [similarly how it's done in Airnode](https://docs.api3.org/reference/airnode/latest/concepts/authorizers.html#special-case-all-authorizers-values-are-empty-arrays) but I think it would be better for this to be explicit and set it to `"all"` instead (string literal instead of an array). Alternatively, we can drop the "whitelist all" use case.

The naming is `"allowedAirnodes"`, because it's about Airnode addresses, but the service is not Airnode but Pusher. I don't mind this, because [Pusher and Airnode are related](https://api3workspace.slack.com/archives/C05S589E7B4/p1695824428354099?thread_ts=1695809464.082009&cid=C05S589E7B4).

The `"allowedAirnodes"` configuration is used for both inserting/querying the data. The "list airnode addresses" endpoint is unchanged, because when all Airnodes are allowed we would need to scan the data storage anyway.

### DIFF
--- a/packages/api/README.md
+++ b/packages/api/README.md
@@ -134,6 +134,25 @@ The port on which the API is served.
 
 The maximum age of the cache header in seconds.
 
+#### `allowedAirnodes`
+
+The list of allowed Airnode addresses. If the list is empty, no Airnode is allowed. To whitelist all Airnodes, set the
+value to `"all"` instead of an array.
+
+Example:
+
+```jsonc
+// Allows pushing signed data from any Airnode.
+"allowedAirnodes": "all"
+```
+
+or
+
+```jsonc
+// Allows pushing signed data only from the specific Airnode.
+"allowedAirnodes": ["0xB47E3D8734780430ee6EfeF3c5407090601Dcd15"]
+```
+
 ## API
 
 The API provides the following endpoints:
--- a/packages/api/README.md
+++ b/packages/api/README.md
@@ -137,13 +137,13 @@ The maximum age of the cache header in seconds.
 #### `allowedAirnodes`
 
 The list of allowed Airnode addresses. If the list is empty, no Airnode is allowed. To whitelist all Airnodes, set the
-value to `"all"` instead of an array.
+value to `"*"` instead of an array.
 
 Example:
 
 ```jsonc
 // Allows pushing signed data from any Airnode.
-"allowedAirnodes": "all"
+"allowedAirnodes": "*"
 ```
 
 or
--- a/packages/api/src/handlers.ts
+++ b/packages/api/src/handlers.ts
@@ -25,12 +25,20 @@ export const batchInsertData = async (requestBody: unknown): Promise<ApiResponse
     );
   }
 
+  // Ensure that the batch of signed that comes from a whitelisted Airnode.
+  const { maxBatchSize, endpoints, allowedAirnodes } = getConfig();
+  if (
+    allowedAirnodes !== 'all' &&
+    !goValidateSchema.data.every((signedData) => allowedAirnodes.includes(signedData.airnode))
+  ) {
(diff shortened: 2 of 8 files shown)


packages/api/src/schema.ts
```diff
@@ -20,6 +24,8 @@ export const endpointsSchema = z
     'Each "urlPath" of an endpoint must be unique'
   );
 
+export const allowedAirnodesSchema = z.union([z.literal('all'), z.array(evmAddressSchema)]);
```
Siegrift, line 27: I am not sure whether not to add `.nonempty` to disallow `[]` since it doesn't make sense to deny all Airnodes.

metobom (commented):
I think it is better to use `all` instead of an empty array.

> I am not sure whether not to add .nonempty to disallow [] since it doesn't make sense to deny all Airnodes.

I think empty arrays should be disallowed.

> The "allowedAirnodes" configuration is used for both inserting/querying the data.

I couldn't catch the "querying" part, can you explain it?

Siegrift:
> I couldn't catch the "querying" part, can you explain it?

I mean that an API call to get data for an non-whitelistested Airnode address with return 403.

packages/api/README.md
```diff
@@ -134,6 +134,25 @@ The port on which the API is served.
 
 The maximum age of the cache header in seconds.
 
+#### `allowedAirnodes`
+
+The list of allowed Airnode addresses. If the list is empty, no Airnode is allowed. To whitelist all Airnodes, set the
```
andreogle, line 139: I agree with `all` as public and `[]` as none 👍🏻

packages/api/src/handlers.ts
```diff
@@ -25,12 +25,20 @@ export const batchInsertData = async (requestBody: unknown): Promise<ApiResponse
     );
   }
 
+  // Ensure that the batch of signed that comes from a whitelisted Airnode.
```
andreogle, line 28: Just to double check I understand, this means that if _any_ of the data come from a non-whitelisted Airnode, the whole batch gets rejected? We don't want to ignore unknown Airnodes and process the rest of the batch?

packages/api/src/schema.ts
```diff
@@ -2,6 +2,10 @@ import { type LogFormat, logFormatOptions, logLevelOptions, type LogLevel } from
 import { uniqBy } from 'lodash';
 import { z } from 'zod';
 
+export const evmAddressSchema = z.string().regex(/^0x[\dA-Fa-f]{40}$/, 'Must be a valid EVM address');
+
+export const evmIdSchema = z.string().regex(/^0x[\dA-Fa-f]{64}$/, 'Must be a valid EVM hash');
```
andreogle, line 7:
```suggestion
export const evmHashSchema = z.string().regex(/^0x[\dA-Fa-f]{64}$/, 'Must be a valid EVM hash');
```
?

andreogle (approved):

packages/api/src/handlers.ts
Siegrift: > Just to double check I understand, this means that if any of the data come from a non-whitelisted Airnode, the whole batch gets rejected?

Yes.

>  We don't want to ignore unknown Airnodes and process the rest of the batch?

In practice the data will be from the same Airnode (pusher) so I think this is fine. And maybe better since if this happens something is probably wrong with the pusher.

packages/api/README.md
Siegrift: Discussed on call to make it `*` instead.

Issue #107, opened 2023-11-03
Improve heartbeat log


Siegrift:
See https://github.com/api3dao/signed-api/pull/100#discussion_r1381757204

PR #108, opened 2023-11-04
Improve heartbeat log

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/107

## Rationale

Rework the log message into a more machine-parseable format. Note, that I've slightly updated how the signing works (we now sign the heartbeat payload object instead of array).

The JSON log message would be:
```json
{"context":{"airnode":"0xbF3137b0a7574563a23a8fC8badC6537F98197CC","configHash":"0x126e768ba244efdb790d63a76821047e163dfc502ace09b2546a93075594c286","currentTimestamp":"1674172803","deploymentTimestamp":"1674172800","nodeVersion":"0.1.0","signature":"0x180b77067b4c319f97fd9b0ff5d6dcbaa02647d0a41497a2256a3b8a42d5f5b16301d1b10fd899c5cba40ea165b4a8fcde7c6e486b21141f5289c0d288c2436f1c","stage":"test"},"level":"info","message":"Sending heartbeat log","ms":"+0ms","timestamp":"2023-01-20T00:00:03.000Z"}
```

You can identify the heartbeat log by the `message` field. The payload will be inside `context` field.

_(You can verify this by changing `jest.spyOn(heartbeatLogger, 'info').mockImplementation();` to `jest.spyOn(heartbeatLogger, 'info')` and running `LOG_FORMAT=json pnpm run test heart` in the heartbeat test)_

### DIFF
--- a/packages/pusher/src/heartbeat/heartbeat.ts
+++ b/packages/pusher/src/heartbeat/heartbeat.ts
@@ -14,11 +14,27 @@ export const initiateHeartbeat = () => {
     if (!goLogHeartbeat.success) logger.error('Failed to log heartbeat', goLogHeartbeat.error);
   }, 1000 * 60); // Frequency is hardcoded to 1 minute.
 };
+export interface HeartbeatPayload {
+  airnode: string;
+  stage: string;
+  nodeVersion: string;
+  currentTimestamp: string;
+  deploymentTimestamp: string;
+  configHash: string;
+  signature: string;
+}
 
-export const signHeartbeat = async (airnodeWallet: ethers.Wallet, heartbeatPayload: unknown[]) => {
+// We need to make sure the object is stringified in the same way every time, so we sort the keys alphabetically.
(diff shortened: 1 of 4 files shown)


packages/pusher/src/heartbeat/heartbeat.ts
```diff
@@ -14,11 +14,27 @@ export const initiateHeartbeat = () => {
...
+  stage: string;
+  nodeVersion: string;
+  currentTimestamp: string;
+  deploymentTimestamp: string;
+  configHash: string;
+  signature: string;
+}
 
-export const signHeartbeat = async (airnodeWallet: ethers.Wallet, heartbeatPayload: unknown[]) => {
+// We need to make sure the object is stringified in the same way every time, so we sort the keys alphabetically.
```
Siegrift, line 27: Note, winston logger sort the context keys alphabetically when formatting the log message. However, objects are unordered so we should not assume any convention and sort these explicitly.

andreogle (approved):

PR #120, opened 2023-11-20
Update SignedApiResponse schema

### LINKED ISSUE #115: Fix signed API response schema
Because the signed API can skip saving some of the beacons it makes sense to also return the number of skipped (ignored) signed data. We do that [here](https://github.com/api3dao/signed-api/blob/0bad6fc8dd6aaffaa12cf099ab6bbf7c98d487c8/packages/api/src/handlers.ts#L116).

The problem is that pusher [response schema is outdated](https://github.com/api3dao/signed-api/blob/0bad6fc8dd6aaffaa12cf099ab6bbf7c98d487c8/packages/pusher/src/validation/schema.ts#L286). This triggers warning in logs.

aquarat:
This PR updates the response schema and associated test resources.

~Maybe having the type on the output payload dictionary is too verbose 🤷, happy to revert of course, but this way it will be obvious in future if there's a mismatch.~

Closes #115

### DIFF
--- a/packages/api/src/handlers.ts
+++ b/packages/api/src/handlers.ts
@@ -1,6 +1,8 @@
 import { go, goSync } from '@api3/promise-utils';
 import { isEmpty, isNil, omit, size } from 'lodash';
 
+import type { SignedApiResponse } from '../../pusher/src/validation/schema';
+
 import { getConfig } from './config';
 import { CACHE_HEADERS, COMMON_HEADERS } from './constants';
 import { deriveBeaconId, recoverSignerAddress } from './evm';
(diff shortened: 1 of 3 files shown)


Siegrift (commented):
👍 LGTM. Is there anything more needed?

> Maybe having the type on the output payload dictionary is too verbose 🤷, happy to revert of course, but this way it will be obvious in future if there's a mismatch.

Not sure what this means.

aquarat:
> 👍 LGTM. Is there anything more needed?
> 
> > Maybe having the type on the output payload dictionary is too verbose 🤷, happy to revert of course, but this way it will be obvious in future if there's a mismatch.
> 
> Not sure what this means.

haha, thanks, if you haven't noticed then it's not a problem 😆 
I'm not sure why the build/CI is failing, I'll work on this now.

Siegrift (approved):
> I originally included the output type in the handler, but by doing so it caused the dist path to shift up 2-levels, so the path became dist/api/src/index.js instead of just dist/index.js. I reverted the change to prevent this from happening.

Yeah, they are separate projects (similarly to airnode packages) so to make that work we would need to have commons package which would export these interfaces for both which is too much work.

I think the interfaces should rarely change so this 👍 LGTM.

PR #121, opened 2023-11-20
Run as non-root users

### LINKED ISSUE #117: Run pusher and signed API as non root docker user
This follows the nodejs best practices.

aquarat:
Closes #117 

As per the linked issue this PR adds largely a copy of the process used in the airseeker-v2 Dockerfile to make the "app" run as a non-privileged user.

I tried to use variable substitution to reduce the duplication but Dockerfiles are seemingly very limited in that area.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,20 +34,40 @@ RUN pnpm run --recursive build
 
 # Create a separate stage for pusher package. We create a temporary stage for deployment and then copy the result into
 # the final stage. Only the production dependencies and package implementation is part of this last stage.
+ENV name="deployed-pusher"
+LABEL application="deployed-pusher" description="Deployed Pusher container"
+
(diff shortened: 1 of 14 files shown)


Dockerfile
```diff
@@ -34,20 +34,40 @@ RUN pnpm run --recursive build
 
 # Create a separate stage for pusher package. We create a temporary stage for deployment and then copy the result into
 # the final stage. Only the production dependencies and package implementation is part of this last stage.
+ENV name="deployed-pusher"
```
Siegrift, line 37: Do we use the ENV (and the similar one below) somewhere?

aquarat: ah, no, I'll remove it. I tried to use this ENV originally to reduce duplication/single source of truth but Dockerfiles seemingly can't work this way.

Issue #84, opened 2023-10-17
Create instructions how to deploy signed API on cloud


Siegrift:
Blocked by https://github.com/api3dao/signed-api/issues/9

We want to have the API deployable to cloud. The deployment would ideally be based on how we deploy pusher.

bbenligiray:
We'll finalize the pusher template and then hopefully this will be just a few additions on top of that

Siegrift:
Added milestone, becuase this issue blocks the performance test (which also has this milestone).

PR #122, opened 2023-11-21
Add CloudFormation template to deploy signed API

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/84

## Rationale

Because we already have a pusher CF template it makes sense to also base the signed API deployment on that. The signed API needs to be accessible from outside, which means it needs more configuration. Notably, we need load balancer and two subnets (AWS requires subnets in at least two zones). I read a bit about best practices and the final result is similar to [this nicely written article](https://containersonaws.com/pattern/public-facing-web-ecs-ec2-cloudformation) (sidenote: I couldn't make their example work).

I verified the deployment by deploying both Pusher and Signed API on AWS, see [this commit](https://github.com/api3dao/signed-api/commit/d91fa91e739eed2798fae3f22ca1f4102ab4054d). The API is accessible [here](http://signed-api-elb-256792608.us-east-1.elb.amazonaws.com/real-time/0xbF3137b0a7574563a23a8fC8badC6537F98197CC) (but will be killed in some time).

### Removal of port

Because we need to configure the load balancer to forward incoming requests to the service we need to know the port number. However, the port number of signed API is configurable. This is arguably confusing because when the service is run inside docker, users can re-publish to any port number on the host machine. Configuring port is then only useful for development server, which we can fix separately. So in short, we hardcode the port number of signed API in docker to port 80.


metobom:
Where do I see the Signed API's URL when I successfully deploy?

packages/api/package.json
```diff
@@ -9,14 +9,14 @@
...
+    "dev": "nodemon --ext ts,js,json,env --exec \"pnpm ts-node src/dev-server.ts\"",
     "docker:build": "docker build --target api --tag api3/signed-api:latest ../../",
-    "docker:start": "docker run --publish 8090:8090 -it --init --volume $(pwd)/config:/app/config --env-file .env --rm api3/signed-api:latest",
+    "docker:run": "docker run --publish 8090:80 -it --init --volume $(pwd)/config:/app/config --env-file .env --rm api3/signed-api:latest",
     "eslint:check": "eslint . --ext .js,.ts --max-warnings 0",
     "eslint:fix": "eslint . --ext .js,.ts --fix",
     "prettier:check": "prettier --check \"./**/*.{js,ts,md,json}\"",
     "prettier:fix": "prettier --write \"./**/*.{js,ts,md,json}\"",
-    "start-prod": "node dist/index.js",
+    "start-prod": "node dist/dev-server.js",
```
andreogle, line 19: I'm a bit confused by the difference between "server" and "dev-server", especially since "start-prod" runs the "**dev**-server" 😬

packages/api/deployment/signed-api-cloudformation-template.json
andreogle, line 1: It's difficult for me to validate this as someone not too familiar with infrastructure best practices

metobom (approved):
I was able to deploy and push data. 
http://signed-api-elb-1361842523.us-east-1.elb.amazonaws.com/real-time/0x3aBC7C59D1d1833cD5973Ce417948523Ba3202D9

packages/api/package.json
Siegrift: Yeah, the naming is not the greatest, but the "start-prod" is actually also a dev-only feature so I think it's fine.

packages/api/deployment/signed-api-cloudformation-template.json
Siegrift: Yeah, I can relate. But by my limited knowledge it should be fine.

Issue #109, opened 2023-11-04
Create signed API and Pusher documentation 


Siegrift:
We already have configuration documented, but what is missing is how the signed API and Pusher works. It would be nice to have a short specification explaining the architecture.

Siegrift:
Also, pushing new docker image should be documented.

Issue #124, opened 2023-11-21
Document release process (versioning and docker release)


Siegrift:
Originally mentioned in https://github.com/api3dao/signed-api/issues/109#issuecomment-1798844770

but for now we do not have proper releases and I use my own account to publish temporary images.

PR #125, opened 2023-11-21
Add spec link

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/109

## Rationale

I wanted to have the high level ideas documented somehere, similarly how it's done with Airseeker. I bookmarked it in Slack and I want to link it in the repo. The services are quite small, but it's the interactions that are important so the document highlights those (e.g. who uses and deploys the services, what calls what, etc...)

The issue also mentions:

> Pushing new docker image should be documented as well.

I've created https://github.com/api3dao/signed-api/issues/124 because we are not published yet (apart from my ad-hoc releases which are specific to me).

> Consider also moving the configuration to a separate README to avoid nested hastag problems.

I don't want to do because it's not an issue and I'd need to update the spec links (which is additional work).

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,10 @@ A monorepo for managing signed data. Consists of:
 - [pusher](./packages/pusher/README.md) - A service for pushing data provider signed data.
 - [e2e](./packages/e2e/README.md) - End to end test utilizing Mock API, pusher and signed API.
 
+Read the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,6 +7,10 @@ A monorepo for managing signed data. Consists of:
 - [pusher](./packages/pusher/README.md) - A service for pushing data provider signed data.
 - [e2e](./packages/e2e/README.md) - End to end test utilizing Mock API, pusher and signed API.
 
+Read the
```
andreogle, line 10: I think this would probably be better as a markdown doc in this repo (`docs/` folder) but it's fine as a Google doc too

andreogle (approved):

PR #127, opened 2023-11-21
Less strict node pnpm requirements

Siegrift:
## Rationale

The `pnpm` is actually getting quite a lot of development and it's annoying because the PRs opened by renovate cannot be merged unless the CI yml file is updated. Both node and pnpm are quite stable so it's better to just force a particular major version of `node` and `pnpm`. 

As a note, we disable the Node.JS because we usually migrate to newer version only when the current nears its EOL.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,8 +2,8 @@
   "name": "signed-api",
   "version": "1.0.0",
   "engines": {
-    "node": "^18.14.0",
-    "pnpm": "^8.10.5"
+    "node": "^18.x",
+    "pnpm": "^8.x"
   },
   "scripts": {
     "build": "pnpm recursive run build",
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "engines": {
     "node": "^18.14.0",
-    "pnpm": "^8.10.2"
+    "pnpm": "^8.10.5"
   },
   "scripts": {
     "build": "pnpm recursive run build",
--- a/renovate.json
+++ b/renovate.json
@@ -3,11 +3,7 @@
   "extends": ["config:base"],
   "packageRules": [
     {
-      "matchPackageNames": ["node"],
-      "enabled": false
-    },
-    {
-      "matchPackageNames": ["ethers"],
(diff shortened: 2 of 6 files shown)


renovate.json
```diff
@@ -3,31 +3,29 @@
...
+      "matchPackageNames": ["ethers", "node"],
       "matchUpdateTypes": ["major"],
       "enabled": false
     },
     {
       "matchDepTypes": ["devDependencies"],
       "matchUpdateTypes": ["patch", "minor"],
       "schedule": ["before 1am on the first day of the month"],
-      "groupName": "devDependencies (non-major)"
+      "groupName": "non-major-devDependencies"
```
Siegrift, line 14: I did this because quotes are a bit annoying when copying the branch name in ZSH.

renovate.json
```diff
@@ -3,31 +3,29 @@
...
-      "groupName": "dependencies (non-major)"
+      "groupName": "non-major-dependencies"
     }
   ],
   "rangeStrategy": "bump",
   "lockFileMaintenance": {
     "enabled": false
   },
   "reviewers": ["Siegrift"],
+  "minimumReleaseAge": "7 days",
```
Siegrift, line 28: I took this from Airnode, inspired by @dcroote . The idea is to only use dependencies that may break. This is only about "feature" updates. Iiuc, the [security updates](https://docs.renovatebot.com/configuration-options/#vulnerabilityalerts) should be opened immediately (but we use dependabot for that now).

andreogle:
Thanks. This was also frustrating for me

package.json
```diff
@@ -2,8 +2,8 @@
   "name": "signed-api",
   "version": "1.0.0",
   "engines": {
-    "node": "^18.14.0",
-    "pnpm": "^8.10.2"
+    "node": "^18.x",
```
Siegrift, line 5: Actually, pnpm doesn't like such syntax and creates:
- https://github.com/api3dao/signed-api/pull/114
- https://github.com/api3dao/signed-api/pull/128

renovate.json
dcroote: Nice, this annoys me too

Issue #123, opened 2023-11-21
Add Loki forwarding to signed API CF template


Siegrift:
The pusher CF template is configured to push the logs to Loki. Signed API should do the same. Ping @bdrhn9 for access if need be.

bbenligiray:
I created a Grafana Cloud account and invited you to it

PR #129, opened 2023-11-22
Forward logs to Loki

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/123

# Rationale

Implement the same Loki integration as is in https://github.com/api3dao/api-integrations/blob/main/data/cloudformation-template.json.

I tested by deploying the template (with few values edited) and Bedirhan confirmed the logs are recieved by Loki.

### DIFF
--- a/packages/signed-api/deployment/cloudformation-template.json
+++ b/packages/signed-api/deployment/cloudformation-template.json
@@ -11,6 +11,21 @@
       "Type": "String",
       "MinLength": 5,
       "Description": "Auth token for Airnode feed."
+    },
+    "grafanaLokiUser": {
+      "Type": "String",
+      "MinLength": 2,
+      "Description": "Username for authenticating Loki API"
+    },
+    "grafanaLokiToken": {
+      "Type": "String",
+      "MinLength": 10,
+      "Description": "Token for authenticating Loki API"
+    },
+    "grafanaLokiEndpoint": {
+      "Type": "String",
+      "MinLength": 10,
+      "Description": "Loki endpoint"
     }
   },
   "Outputs": {
@@ -40,6 +55,31 @@
         "RequiresCompatibilities": ["FARGATE"],
         "ExecutionRoleArn": { "Ref": "ECSTaskRole" },
         "ContainerDefinitions": [
+          {
+            "Essential": true,
+            "Image": "grafana/fluent-bit-plugin-loki:2.9.1-amd64",
+            "Name": "SignedApiLogForwarder",
+            "FirelensConfiguration": {
+              "Type": "fluentbit",
+              "Options": {
+                "enable-ecs-log-metadata": "true"
+              }
+            },
+            "LogConfiguration": {
+              "LogDriver": "awslogs",
+              "Options": {
+                "awslogs-group": {
+                  "Ref": "SignedApiLogsGroup"
+                },
+                "awslogs-region": {
+                  "Ref": "AWS::Region"
+                },
+                "awslogs-create-group": "true",
+                "awslogs-stream-prefix": "firelens"
+              }
+            },
+            "MemoryReservation": 50
+          },
           {
             "Name": "signed-api-container",
             "Image": "<DOCKER_IMAGE>",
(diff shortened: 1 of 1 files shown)


packages/api/deployment/signed-api-cloudformation-template.json
```diff
@@ -53,11 +94,33 @@
...
+                      },
+                      "@",
+                      {
+                        "Ref": "grafanaLokiEndpoint"
+                      },
+                      "/loki/api/v1/push"
+                    ]
+                  ]
+                },
+                "Labels": "{app=\"signed-api\",signedApiId=\"<SIGNED_API_ID>\"}",
```
Siegrift, line 120: As a note, this will not work as of now. The endpoint requires the following schema:
```
export const generateTokenInputSchema = z
  .object({
    airnode: evmAddressSchema
  })
  .strict();
```

but that needs to be changed. Leaving that to @bdrhn9 .

packages/api/deployment/signed-api-cloudformation-template.json
```diff
@@ -14,20 +31,44 @@
...
         "Cpu": "256",
         "Memory": "512",
         "NetworkMode": "awsvpc",
         "RequiresCompatibilities": ["FARGATE"],
         "ExecutionRoleArn": { "Ref": "ECSTaskRole" },
         "ContainerDefinitions": [
+          {
+            "Essential": true,
+            "Image": "grafana/fluent-bit-plugin-loki:2.9.1-amd64",
+            "Name": "SignedApiLogForwarder-<SOME_ID>",
```
Siegrift, line 50: This probably doesn't need `-<SOME_ID>` suffix.

packages/api/deployment/signed-api-cloudformation-template.json
bdrhn9: We chose the airnode address as a good differentiator among providers. However, it doesn't solve our problem here because the signed API is not tied to any mnemonic. In this case, using a pseudo airnode address like 0x000...03 could be an option. I would like to note that the tokens required for Loki can only be generated by @metobom

packages/api/deployment/signed-api-cloudformation-template.json
bdrhn9: I'd rather to not use too.

packages/api/deployment/signed-api-cloudformation-template.json
Siegrift: Mhh, I am not familiar with how the integration works, but we can't we have different schemas for different services? We will also deploy Airseeekers and some other services in the future and I guess we would like to forward the logs from each. Using special Airnode addresses will not be maintainable.

bdrhn9: I spoke with @bbenligiray , and I'll try to forward them to you. We don't expect API partners to deploy their signed-api using Cloudformation, and we also don't expect them to push their signed-api logs to our centralized log instance. Therefore, this template is only to be used by API3 staff responsible for deploying signed-api to AWS instances. For our purposes, as there will be more than one deployment, it's still necessary to have a label to differentiate them. So, as you purposed `signedApiId`, `deploymentId`, or `stage` can be options.

Siegrift: > We don't expect API partners to deploy their signed-api using Cloudformation, and we also don't expect them to push their signed-api logs to our centralized log instance.

Mhh, how will they do the deployment? What is the reasoning for not forwarding Signed API logs to us (I think the reasoning for forwarding Pusher logs is similar for Signed API)?

packages/signed-api/deployment/cloudformation-template.json
```diff
@@ -86,11 +126,33 @@
...
+                      },
+                      "@",
+                      {
+                        "Ref": "grafanaLokiEndpoint"
+                      },
+                      "/loki/api/v1/push"
+                    ]
+                  ]
+                },
+                "Labels": "{app=\"signed-api\",airnodeAddress=\"<AIRNODE_ADDRESS>\"}",
```
Siegrift, line 152: I now have access to Grafana Loki and I see we can filter the combination of both `app` and `airnodeAddress`. As we plan to deploy separate Signed API per Airnode, the `airnodeAddress` makes sense to me.

As a bonus this works fine with the already created endpoint for Loki tokens creation.

bdrhn9: In [cloudformation template](https://github.com/api3dao/api-integrations/blob/main/data/cloudformation-template.json#L182) for `airnode-feed`, `airnode` is utilized instead of `airnodeAddress`. It would be better have same label for both.

bdrhn9 (approved):
LGTM 👍🏼

packages/signed-api/deployment/cloudformation-template.json
Siegrift: Good point. Thanks.

Issue #80, opened 2023-10-11
Avoid changing secrets in e2e tests


Siegrift:
The https://github.com/api3dao/signed-api/blob/33e59702d55a0a9033d5a861f62bf4efa89cffbf/packages/e2e/src/pusher/secrets.env#L3C1-L3C1 is pretty annoying and error prone. It would be nice to just set it once.

We can make the file an example and let the user create a new one with correct ENV value set. This will also fail, when user does not create the file.

PR #136, opened 2023-11-25
Avoid changing secrets in e2e tests

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/80

### DIFF
--- a/packages/e2e/src/pusher/secrets.example.env
+++ b/packages/e2e/src/pusher/secrets.example.env
@@ -1,0 +1,5 @@
+# Secrets must NOT be quoted.
+#
+# Use "host.docker.internal" if using Docker for Desktop, otherwise laeve it set to "localhost".
+SIGNED_API_URL=http://localhost:8090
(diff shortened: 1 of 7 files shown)


packages/e2e/src/pusher/secrets.example.env
```diff
@@ -0,0 +1,5 @@
+# Secrets must NOT be quoted.
```
andreogle, line 1: Why is this?

andreogle (approved):

packages/e2e/src/pusher/secrets.example.env
Siegrift: This [was a thing in Airnode](https://github.com/api3dao/airnode/issues/1698), but it was because of how the secrets were processed in Terraform and its probably fixed there as well.

I've double check quoting an unquoting secrets in when running Pusher with/without docker and it works fine. I'll remove these messages.

PR #140, opened 2023-11-28
Implement various minor fixes

Siegrift:
## Rationale

@andreogle found various issues when running signed API and Airseeker locally. Specifically:
1. `DEV_SERVER_PORT` not working
2.  `DEV_SERVER_PORT` is not set to 8090 by default (pusher expects signed API on this port)
3. Add more context to `Unauthorized Airnode address` message.
4. Default configuration of signed API has `allowedAirnodes` set to random Airnode instead of the address of the pusher configured for development (with fixed mnemonic).

These are all implemented, and I've slightly changed the function for generating error response.

There was another point, whether pusher needs to use both `.env` and `.secrets.env` but there is a conceptual difference between those so I'm leaving that in place.

### DIFF
--- a/packages/api/src/dev-server.ts
+++ b/packages/api/src/dev-server.ts
@@ -4,7 +4,7 @@ import { fetchAndCacheConfig } from './config';
 import { logger } from './logger';
 import { DEFAULT_PORT, startServer } from './server';
 
-const portSchema = z.number().int().positive();
+const portSchema = z.coerce.number().int().positive();
 
 const startDevServer = async () => {
   const config = await fetchAndCacheConfig();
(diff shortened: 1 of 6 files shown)


packages/api/src/dev-server.ts
```diff
@@ -4,7 +4,7 @@ import { fetchAndCacheConfig } from './config';
 import { logger } from './logger';
 import { DEFAULT_PORT, startServer } from './server';
 
-const portSchema = z.number().int().positive();
+const portSchema = z.coerce.number().int().positive();
```
andreogle, line 7: Nice - I wasn't aware of `coerce` before 👍🏻

packages/api/src/utils.ts
```diff
@@ -15,8 +15,11 @@ export const isIgnored = (signedData: SignedData, ignoreAfterTimestamp: number)
 export const generateErrorResponse = (
   statusCode: number,
   message: string,
-  detail?: string,
-  extra?: unknown
+  context?: Record<string, unknown>
```
andreogle, line 18: 👍🏻

andreogle (approved):

Issue #144, opened 2023-11-30
Improve consistency of log statements


Siegrift:
All logs should end with a dot.

PR #148, opened 2023-12-01
Improve log style consistency

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/144

### DIFF
--- a/packages/airnode-feed/src/heartbeat/heartbeat.ts
+++ b/packages/airnode-feed/src/heartbeat/heartbeat.ts
@@ -7,6 +7,10 @@ import { loadRawConfig } from '../validation/config';
 
 import { heartbeatLogger } from './logger';
 
+// Intentionally making the message as constant so that it is not accidentally changed. API integrations team (and maybe
+// other teams, such as monitoring) will listen for this exact message to parse the heartbeat.
+const HEARTBEAT_LOG_MESSAGE = 'Sending heartbeat log.';
+
 export const initiateHeartbeat = () => {
   logger.debug('Initiating heartbeat loop.');
   setInterval(async () => {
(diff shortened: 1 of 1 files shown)


packages/airnode-feed/src/heartbeat/heartbeat.ts
```diff
@@ -7,11 +7,15 @@ import { loadRawConfig } from '../validation/config';
 
 import { heartbeatLogger } from './logger';
 
+// Intentionally making the message as constant so that it is not accidentally changed. API integrations team (and maybe
+// other teams, such as monitoring) will listen for this exact message to parse the heartbeat.
+const HEARTBEAT_LOG_MESSAGE = 'Sending heartbeat log.';
```
Siegrift, line 12: cc: @bdrhn9, @metobom  I've changed the log message (there is not a dot at the end).

Siegrift:
> Kinda puzzled by this one. What's the reason for the full stops everywhere?

I dislike the inconsistency between some messages having it and some don't. We could do with or without a dot and I chose to add it because sometimes you want to write two sentences and use a dot in the middle. E.g. `Some error happened. Skipping update.`. It feels wierd where is no dot at the end but there is one in the middle.

bbenligiray:
I omit the period if I'm writing a single sentence and use it otherwise. It's more of a style thing though, I find single sentences ending with periods too sassy.

Siegrift:
Yeah, it's mostly matter of style and it's not a big deal if it's inconsistent, but I still consider it an improvement. I did a quick search and there are people [recommending the same log style](https://stackoverflow.com/questions/1136829/do-you-end-your-exception-messages-with-a-period).

Issue #151, opened 2023-12-01
Fix "PayloadTooLargeError: request entity too large"


Siegrift:
I encountered this error when attempting a performance test:
```
PayloadTooLargeError: request entity too large
    at readStream (/app/node_modules/.pnpm/raw-body@2.5.1/node_modules/raw-body/index.js:156:17)
    at getRawBody (/app/node_modules/.pnpm/raw-body@2.5.1/node_modules/raw-body/index.js:109:12)
    at read (/app/node_modules/.pnpm/body-parser@1.20.1/node_modules/body-parser/lib/read.js:79:3)
    at jsonParser (/app/node_modules/.pnpm/body-parser@1.20.1/node_modules/body-parser/lib/types/json.js:135:5)
    at Layer.handle [as handle_request] (/app/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/app/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/router/index.js:328:13)
    at /app/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/router/index.js:286:9
    at Function.process_params (/app/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/router/index.js:346:12)
    at next (/app/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/router/index.js:280:10)
    at expressInit (/app/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/middleware/init.js:40:5)
```

The issue is the default limit on payload size in express json parser which is set to 100kb which is arguably too low. The express middleware errors are also not handled well. We should attach error handler middleware and process these errors. See:
* https://stackoverflow.com/questions/19917401/error-request-entity-too-large
* https://masteringjs.io/tutorials/express/req#:~:text=By%20default%2C%20express.,the%20limit%20option%20to%20express.
* https://expressjs.com/en/guide/error-handling.html

PR #152, opened 2023-12-01
Increase payload size, add default express error handler

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/151

## Rationale

I've set the maximum payload limit to 10mb which should be more then enough for practical use cases. The higher the limit is the bigger is the risk of Signed API being spammed by attackers. We have Airnode allowlist in the config, but even processing such huge payloads incurs some overhead.

I've also added an e2e test that validates this.

### DIFF
--- a/packages/api/src/server.ts
+++ b/packages/api/src/server.ts
@@ -12,7 +12,8 @@ export const DEFAULT_PORT = 80;
 export const startServer = (config: Config, port: number) => {
   const app = express();
 
-  app.use(express.json());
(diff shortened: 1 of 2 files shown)


Siegrift:
Thanks. For full context the signed data is roughly ~400 bytes, so for 100kb we could do 100000 / 400 = 250, with 10mb we can do 10000000 / 400 = 25000. I think 1mb would also suffice for practical purposes, but I'd leave it set to 10mb.

andreogle (approved):

Issue #149, opened 2023-12-01
Add stage to Signed API


Siegrift:
Third point of https://api3workspace.slack.com/archives/C05S589E7B4/p1700498242003409

In https://github.com/api3dao/signed-api/pull/99 we've added the stage for Pusher, but the PR had confusing title. The stage was not added to Signed API (but it should).

PR #156, opened 2023-12-04
Add stage to Signed API

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/149


Siegrift:
Actually, we should probably add also the version (similar to nodeVersion in Airnode).

andreogle (approved):

Issue #158, opened 2023-12-06
Push data immediately as they are fetched


Siegrift:
See: https://api3workspace.slack.com/archives/C05S589E7B4/p1701869882188689?thread_ts=1701869036.467659&cid=C05S589E7B4

PR #159, opened 2023-12-06
Performance test fixes and optimizations

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/161
Closes https://github.com/api3dao/signed-api/issues/158

## Rationale

I cherry picked various fixes from https://github.com/api3dao/signed-api/pull/154 and opening a separate PR. Review commit by commit.


packages/api/src/server.ts
```diff
@@ -16,7 +16,7 @@ export const startServer = (config: Config, port: number) => {
   app.use(express.json({ limit: '10mb' }));
 
   app.post('/', async (req, res) => {
-    logger.info('Received request "POST /".', req.body);
+    logger.debug('Received request "POST /".', req.body);
```
andreogle, line 19: Not sure I agree with making these debug

andreogle (approved):
👍🏻

packages/api/src/server.ts
Siegrift: I changed it to debug, because the request body can be huge (up the point that it's trimmed by the logger). I can make a `logger.info` call that just logs that a request has been made as a compromise?

Or are you thinking about something else?

bbenligiray: Related to that @metobom wanted to hack this into Airnode feed processing snippets to debug API integration issues (https://github.com/api3dao/tasks/issues/400#issuecomment-1850355602) I told him not to because we don't know the API response size.

andreogle: > because the request body can be huge

Ah right. Then I kinda like making the request (without params) an `info` level log and the params `debug`

PR #168, opened 2023-12-13
Add additional signed data checks

### LINKED ISSUE #162: Do not store unusable data in Signed API
[Here](https://github.com/api3dao/signed-api/blob/52d03eb00205473dfbfbcfa729fda623d1825ba2/packages/api/src/handlers.ts#L60) we do a check whether the signed data is new (and shall be added to state). This does not check whether the signed data isn't too old. Imagine this case:
1. There is a signed data for a beacon and all GET endpoints can return this value (it respects the delays of each).
2. We push a historical (too old) signed data.
3. This check will ignore all beacons and consider this data as "new" and adds it to state.
4. We immediatelly prune it [here](https://github.com/api3dao/signed-api/blob/52d03eb00205473dfbfbcfa729fda623d1825ba2/packages/api/src/handlers.ts#L78) but the endpoint considers the data as "added".
5. The endpoint [response count](https://github.com/api3dao/signed-api/blob/52d03eb00205473dfbfbcfa729fda623d1825ba2/packages/api/src/handlers.ts#L87) is incorrect

aquarat:
Closes #162 

This PR expands the existing signed data freshness check to confirm that the sample we've received is newer than what we already have (if we have something to compare against).

It also checks that the received sample is not older than an hour.

I'm not sure if this satisfies the issue? 🤷

### DIFF
--- a/packages/api/src/handlers.ts
+++ b/packages/api/src/handlers.ts
@@ -87,7 +87,7 @@ export const batchInsertData = async (requestBody: unknown): Promise<ApiResponse
     const goReadDb = await go(async () => get(signedData.airnode, signedData.templateId, requestTimestamp));
 
     if (goReadDb.data && requestTimestamp <= Number.parseInt(goReadDb.data.timestamp, 10)) {
-      logger.debug('Not storing signed data because the signed data is not newer than the db sample.', {
+      logger.debug('Not storing signed data because the signed data is not newer than what we already have.', {
         signedData,
       });
       continue;
--- a/packages/api/src/handlers.ts
+++ b/packages/api/src/handlers.ts
@@ -85,8 +85,16 @@ export const batchInsertData = async (requestBody: unknown): Promise<ApiResponse
   for (const signedData of batchSignedData) {
     const requestTimestamp = Number.parseInt(signedData.timestamp, 10);
     const goReadDb = await go(async () => get(signedData.airnode, signedData.templateId, requestTimestamp));
-    if (goReadDb.data && requestTimestamp === Number.parseInt(goReadDb.data.timestamp, 10)) {
-      logger.debug('Not storing signed data because signed data with the same timestamp already exists.', {
+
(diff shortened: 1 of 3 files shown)


packages/api/src/handlers.ts
```diff
@@ -84,9 +84,17 @@ export const batchInsertData = async (requestBody: unknown): Promise<ApiResponse
   // is acceptable, but we only want to store one data for each timestamp.
   for (const signedData of batchSignedData) {
     const requestTimestamp = Number.parseInt(signedData.timestamp, 10);
-    const goReadDb = await go(async () => get(signedData.airnode, signedData.templateId, requestTimestamp));
-    if (goReadDb.data && requestTimestamp === Number.parseInt(goReadDb.data.timestamp, 10)) {
-      logger.debug('Not storing signed data because signed data with the same timestamp already exists.', {
+    const goReadDb = await go(async () => get(signedData.airnode, signedData.templateId));
+
+    if (goReadDb.data && requestTimestamp <= Number.parseInt(goReadDb.data.timestamp, 10)) {
```
Siegrift, line 89: We want to store even older data we have because it might be served by other (delayed) endpoint. To address the issue we would need to check whether the signed data will be pruned, but checking that is more complex than I initially thought (and probably not worth relative to the added benefit).

Siegrift (commented):
As I mentioned in the comment, I think fixing the confusing case mentioned in the issue is not worth it. It won't happen in practice and I only reproduced it because I saved some very old signed data response in my Postman.

WDYT?

aquarat:
> As I mentioned in the comment, I think fixing the confusing case mentioned in the issue is not worth it. It won't happen in practice and I only reproduced it because I saved some very old signed data response in my Postman.
> 
> WDYT?

Sounds good to me, we can always come back to it. In practice we shouldn't be pushing very old signed data to the chain anyway (ultimately that's where the data will be going via Airseeker).

Issue #22, opened 2023-09-20
Implement API authentication


Siegrift:
See: https://api3workspace.slack.com/archives/C05S589E7B4/p1695139756669989?thread_ts=1695136091.683809&cid=C05S589E7B4

At least the non-delayed data should be hidden behind auth.

bbenligiray:
In the self-hosted setup the API provider may want to use their own authentication scheme, which is why I said this feels out of scope similar to caching

Siegrift:
Based on [this](https://api3workspace.slack.com/archives/C05S589E7B4/p1701869036467659) discussion, we want to use the API authentication to avoid the need to verify signed data, which turns out to be a big bottlenect.

Siegrift:
The outcome was to have the signed API POST request bodies to be signed. The signer has to be in [the whitelist](https://github.com/api3dao/signed-api/issues/98) and the match Airnode addresses in the body.

PR #174, opened 2023-12-20
Implement bearer token auth

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/22

## Rationale

The Signed API configuration changes are inspired by the following idea:
```
{
  "endpoints": [
    {
      "urlPath": "/default",
    -> "authTokens": ["token"],
      "delaySeconds": 0
    },
    {
      "urlPath": "/delayed",
   -> "authTokens": null,
      "delaySeconds": 15
    }
  ],
  "allowedAirnodes": [
    {
      "address": "0xbF3137b0a7574563a23a8fC8badC6537F98197CC",
   -> "authTokens": ["airnode-token-1", "airnode-token-2"]
    }
  ],
  "stage": "local",
  "version": "0.1.0"
}
```

Each endpoint can have Bearer token authentication. Similarly for the endpoint to push the data from Airnode feed. Read the configuration changes (READMEs) and the changes in the example configuration files.

Related
- [x] Readme config changes
- [x] Update spec (high level reasoning why we use auth)

### DIFF
--- a/packages/airnode-feed/README.md
+++ b/packages/airnode-feed/README.md
@@ -283,11 +283,12 @@ Configuration for the signed APIs. Each signed API is defined by a `signedApiNam
 example:
 
 ```jsonc
-// Defines a single signed API.
+// Defines a single signed API that uses AUTH_TOKEN secret as Bearer token when pushing signed data to signed API.
 "signedApis": [
   {
     "name": "localhost",
-    "url": "http://localhost:8090"
+    "url": "http://localhost:8090",
+    "authToken": "${AUTH_TOKEN}"
   }
 ]
 ```
@@ -303,6 +304,13 @@ The name of the signed API.
 ###### `url`
 
 The URL of the signed API.
+
+#### `authToken`
+
+The authentication token used to authenticate with the signed API. It is recommended to interpolate this value from
+secrets.
+
+If the signed API does not require authentication, set this value to `null`.
 
 #### `ois`
 
--- a/packages/api/README.md
+++ b/packages/api/README.md
@@ -145,15 +145,17 @@ The API needs to be configured with endpoints to be served. This is done via the
 ```jsonc
 // Defines two endpoints.
 "endpoints": [
-  // Serves the non-delayed data on URL path "/real-time".
+  // Serves the non-delayed data on URL path "/real-time". Requesters need to provide the "some-secret-token" as Bearer token.
   {
     "urlPath": "/real-time",
-    "delaySeconds": 0
+    "delaySeconds": 0,
+    "authTokens": ["some-secret-token"],
   },
-  // Serves the data delayed by 15 seconds on URL path "/delayed".
+  // Serves the data delayed by 15 seconds on URL path "/delayed". No authentication is required.
   {
     "urlPath": "/delayed",
-    "delaySeconds": 15
+    "delaySeconds": 15,
+    "authTokens": null,
   }
 ]
 ```
@@ -170,6 +172,14 @@ dashes.
 ###### `delaySeconds`
 
 The delay in seconds for the endpoint. The endpoint will only serve data that is older than the delay.
+
+###### `authTokens`
+
+The nonempty list of
+[Bearer authentication tokens](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#bearer) allowed to query
+the data.
+
+In case the endpoint should be publicly available, set the value to `null`.
 
 #### `cache` _(optional)_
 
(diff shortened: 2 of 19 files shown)


packages/airnode-feed/README.md
```diff
@@ -304,6 +305,13 @@ The name of the signed API.
 
 The URL of the signed API.
 
+#### `authToken`
+
+The authentication token used to authenticate with the signed API. It is recommended to interpolate this value from
+secrets.
+
+If the signed API does not require authentication, set this value to `null`.
```
andreogle, line 313: Any reason for making this field required and having it be `null` as opposed to just making the entire field optional?

packages/api/README.md
```diff
@@ -145,15 +145,17 @@ The API needs to be configured with endpoints to be served. This is done via the
...
+    "delaySeconds": 0,
+    "authTokens": ["some-secret-token"],
   },
-  // Serves the data delayed by 15 seconds on URL path "/delayed".
+  // Serves the data delayed by 15 seconds on URL path "/delayed". No authentication is required.
   {
     "urlPath": "/delayed",
-    "delaySeconds": 15
+    "delaySeconds": 15,
+    "authTokens": null,
```
andreogle, line 158: Same with this - nullable vs optional

packages/api/README.md
```diff
@@ -145,15 +145,17 @@ The API needs to be configured with endpoints to be served. This is done via the
 ```jsonc
 // Defines two endpoints.
 "endpoints": [
-  // Serves the non-delayed data on URL path "/real-time".
+  // Serves the non-delayed data on URL path "/real-time". Requesters need to provide the "some-secret-token" as Bearer token.
```
andreogle, line 148: I assume the user only needs to pick one of the tokens to use from the list

packages/api/README.md
```diff
@@ -171,6 +173,14 @@ dashes.
 
 The delay in seconds for the endpoint. The endpoint will only serve data that is older than the delay.
 
+###### `authTokens`
+
+The nonempty list of
```
andreogle, line 178: I must have missed this somewhere, but what's the reasoning for making this a list?

packages/api/src/handlers.ts
```diff
@@ -4,39 +4,58 @@ import { isEmpty, isNil, omit } from 'lodash';
...
+    // Find the allowed airnode and extract the request token.
+    const allowedAirnode = allowedAirnodes.find((allowedAirnode) => allowedAirnode.address === airnodeAddress);
+    const authToken = extractBearerToken(authorizationHeader);
+
+    // Check if the airnode is allowed and if the auth token is valid.
+    const isAirnodeAllowed = Boolean(allowedAirnode);
+    const isAuthTokenValid = allowedAirnode?.authTokens === null || allowedAirnode?.authTokens.includes(authToken!);
+    if (!isAirnodeAllowed || !isAuthTokenValid) {
+      if (isAirnodeAllowed) {
+        logger.debug(`Invalid auth token`, { allowedAirnode, authToken });
```
andreogle, line 53: Shouldn't this be a `warn` instead?

andreogle (approved):

packages/airnode-feed/README.md
Siegrift: In general, I like being explicit. Having no auth can be a critical mistake if omitted by accident. Having it explicitly set to `null` makes this decision explicit (and is visible to anyone reading the config).

For example, the [cache](https://github.com/api3dao/signed-api/blob/6f1d0a06bc759a7388fb089e3e36edcf8a497c53/packages/api/src/schema.ts#L38) section is optional, because no cache is a reasonable default. But arguably, that field could be also made explicit.

packages/api/README.md
Siegrift: Same reason as in https://github.com/api3dao/signed-api/pull/174#discussion_r1433221401. Let's discuss there.

packages/api/README.md
Siegrift: Yeah, the array is in case we want to hand out different tokens to different parties. That makes it possible to revoke a specific token and leave the rest.

packages/api/README.md
Siegrift: It's an array in case we want to hand out different tokens to different parties. That makes it possible to revoke a specific token and leave the rest.

packages/api/src/handlers.ts
Siegrift: It is probably because someone may be spamming the Signed API and in that case we would end up with many of these logs.

Issue #167, opened 2023-12-13
Publish the packages on NPM


Siegrift:
See: https://api3workspace.slack.com/archives/C05S589E7B4/p1702420010173829

The packages should be released with the same version and as `@api3/signed-api` and `@api3/airnode-feed`.

Issue #165, opened 2023-12-12
Fix release instructions


Siegrift:
[The release instructions](https://github.com/api3dao/signed-api/blob/883ec16fb0dec4613e37bdb1e0d5d2bf08d23784/packages/api/README.md?plain=1#L234) were based on the commons repo, but because signed-api is a monorepo running `pnpm version` actually does not create git commit nor git tag. 

Also, we should generate a GH release from the tag. As a note, if we create git tags manually, the `--follow-tags` only pushes annotated tags.

Issue #164, opened 2023-12-11
Document Docker release instuctions on macOS


Siegrift:
It's necessary to build the image for platform `linux/amd64` in oder to dpeloy on AWS. This can be done on macOS with `buildx`. E.g.:

```sh
docker buildx build --platform linux/amd64 --target api --tag api3/signed-api:latest ../../
```

similarly for Airnode feed.

PR #179, opened 2023-12-21
Automate release process, support NPM publishing

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/164
Closes https://github.com/api3dao/signed-api/issues/165
Closes https://github.com/api3dao/signed-api/issues/167

## Rationale

We agreed on call to publish the packages with the same version (similarly to Airnode). I looked for potential solutions that do this, because such workspace versioning not natively supported by PNPM nor NPM. We can use changesets (what Airnode does) or do this ourselves. I did a small pros/cons:
- Changesets
  - (-) Need to write changesets manually
  - (-) Process to follow
  - (+) Bumps version and does release for projects automatically
  - (+) When done by CI it generates Github release notes per project
  - (+) Generated changelog for each project
  - (+) Tested by many projects in the wild

- Custom Script
  - (-) Code and maintenance overhead
  - (+) Simpler process to follow
  - (+) Single git tag for both projects
  - (+-) Common Github release for both project generated from PR titles
  
  and ended up with implementing a version script.
  
  ### Related

* I've renamed `api` workspace to `signed-api` so that the package is the same as on NPM.
* Published packages are `@api3/signed-api` and `@api3/airnode-feed`.
* I've tested the publishing locally with Verdaccio. The imports looked good, but I only exported a limited functionality (mostly configuration utils and schemas).


package.json
```diff
@@ -1,19 +1,24 @@
...
+    "node": ">=18.19.0",
     "pnpm": "^8.12.0"
   },
   "scripts": {
     "build": "pnpm recursive run build",
     "clean": "pnpm recursive run clean",
-    "docker:build": "pnpm recursive run docker:build",
+    "create-docker-release": "ts-node scripts/create-docker-release.ts",
+    "create-npm-release": "ts-node scripts/create-npm-release.ts",
+    "docker:build": "pnpm recursive run --sequential docker:build",
```
Siegrift, line 13: `--sequential` works better because the second Docker image can reuse the stage built by the first one.

packages/airnode-feed/src/heartbeat/heartbeat.test.ts
```diff
@@ -41,7 +32,16 @@ describe(logHeartbeat.name, () => {
 
     await logHeartbeat();
 
-    expect(heartbeatLogger.info).toHaveBeenCalledWith('Sending heartbeat log.', expectedLogMessage);
+    expect(heartbeatLogger.info).toHaveBeenCalledWith(
```
Siegrift, line 35: It was annoying that the test needed to be updated after every config change, especially after a new version was created.

package.json
andreogle: Nice 👍🏻

package.json
```diff
@@ -1,19 +1,24 @@
...
   "engines": {
-    "node": "^18.19.0",
+    "node": ">=18.19.0",
     "pnpm": "^8.12.0"
   },
   "scripts": {
     "build": "pnpm recursive run build",
     "clean": "pnpm recursive run clean",
-    "docker:build": "pnpm recursive run docker:build",
+    "create-docker-release": "ts-node scripts/create-docker-release.ts",
```
andreogle, line 11: Nit: I think `release: docker` and `release:npm` would be easier to find in the list here and are more consistent with other script names. I also think the script files should match (`scripts/release-docker.ts`)

packages/airnode-feed/src/npm-exports.ts
```diff
@@ -0,0 +1,5 @@
+// NOTE: This file defines the NPM exports. Everything that should be available to import from the NPM package should be
```
andreogle, line 1: Isn't it more common to just export from `index.ts`?

andreogle (approved):

package.json
Siegrift: Maybe `create-release:docker` and `create-release:npm` because the scripts do not release anything on their own?

packages/airnode-feed/src/npm-exports.ts
Siegrift: I thought about that, but I'd need to change all Docker entrypoints and package.json scripts which feels like unnecessary risk of breaking things.

package.json
andreogle: Ok that works

Siegrift: I renamed the package scripts, but left the script names intact because I didn't like using `:` in the file name.

PR #189, opened 2024-01-04
Add airnode address to POST endpoint as a path parameter

### LINKED ISSUE #185: Add airnode address to `POST /` endpoint
When we try to scale signed-api over airnode addresses on a single domain, it's possible to route `GET /delayed/<airnodeAddress>` and `GET /real-time/<airnodeAddress>` because the airnode address uniquely determines which container to use. However, routing the `POST /` address that accepts signed data complicates the routing process in a simple way. So adding airnode address to POST endpoint e.g. `POST /<airnodeAddress>` will make routing easier. 

Blocker for https://github.com/api3dao/tasks/issues/507

bdrhn9:
Closes #185

### DIFF
--- a/packages/airnode-feed/src/api-requests/signed-api.ts
+++ b/packages/airnode-feed/src/api-requests/signed-api.ts
@@ -45,7 +45,7 @@ export const pushSignedData = async (group: SignedApiUpdate) => {
     const signedApi = signedApis.find((a) => a.name === signedApiName)!;
     const goAxiosRequest = await go<Promise<unknown>, AxiosError>(async () => {
       logger.debug('Posting batch payload.', { batchPayload });
-      const axiosResponse = await axios.post(signedApi.url, batchPayload, {
+      const axiosResponse = await axios.post(new URL(airnode, signedApi.url).href, batchPayload, {
         headers: {
           'Content-Type': 'application/json',
           ...(signedApi.authToken ? { Authorization: `Bearer ${signedApi.authToken}` } : {}),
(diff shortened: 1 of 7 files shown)


packages/airnode-feed/src/api-requests/signed-api.ts
```diff
@@ -45,7 +45,7 @@ export const pushSignedData = async (group: SignedApiUpdate) => {
     const signedApi = signedApis.find((a) => a.name === signedApiName)!;
     const goAxiosRequest = await go<Promise<unknown>, AxiosError>(async () => {
       logger.debug('Posting batch payload.', { batchPayload });
-      const axiosResponse = await axios.post(signedApi.url, batchPayload, {
+      const axiosResponse = await axios.post(new URL(airnode, signedApi.url).href, batchPayload, {
```
bdrhn9, line 48: I didn't prefer to use just template strings here. If I create template string like `${signedApi.url}/${airnode}` and signedApi.url includes `/` at the end of the string, it'll result in something like `https://signed-api.bla.com//0x00.abc`. So I preferred safe way to join path and base url.

packages/signed-api/src/handlers.ts
```diff
@@ -22,8 +22,27 @@ import {
 // important for the delayed endpoint which may not be allowed to return the fresh data yet.
 export const batchInsertData = async (
   authorizationHeader: string | undefined,
-  requestBody: unknown
+  requestBody: unknown,
+  airnodeAddress: string
 ): Promise<ApiResponse> => {
+  // Ensure that the batch of signed that comes from a whitelisted Airnode.
```
bdrhn9, line 28: Here I moved checking authentication at the beginning of the handler because we now have `airnodeAddress` in advance (previously we need to parse and validate request body).

Siegrift (approved):
👍 LGTM

bbenligiray:
README is outdated, for example https://github.com/api3dao/signed-api/blob/main/packages/signed-api/README.md#api

Siegrift:
Thanks. Will look into it.

Issue #186, opened 2024-01-02
Investigate 0.2.0 docker images issue


Siegrift:
@metobom reported an issue when trying to deploy Docker image for 0.2.0 Signed API:

```
Node.js v18.19.0
node:fs:596
  handleErrorFromBinding(ctx);
  ^

Error: ENOENT: no such file or directory, open '/app/config/signed-api.json'
    at Object.openSync (node:fs:596:3)
    at readFileSync (node:fs:464:35)
    at loadConfigFromFilesystem (/app/dist/src/config.js:30:78)
    at fetchConfig (/app/dist/src/config.js:37:57)
    at fetchAndCacheConfig (/app/dist/src/config.js:20:30)
    at main (/app/dist/src/index.js:7:59)
    at Object.<anonymous> (/app/dist/src/index.js:11:6)
    at Module._compile (node:internal/modules/cjs/loader:1356:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1414:10)
    at Module.load (node:internal/modules/cjs/loader:1197:32) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/app/config/signed-api.json'
}
```

metobom:
I got this error while trying to deploy Airnode feed 0.2.0. Please see `IMAGE LAYERS` in here https://hub.docker.com/layers/api3/airnode-feed/0.2.0/images/sha256-2f5d179d0f65acb85bf9c5ef7a20fd034ed4d14abf7d26ce40f5f5181f9cad07?context=explore

bdrhn9:
I think I found the cause of the issue. In the following package.json script, `--target` is missing for the script used to build docker images on macOS.

https://github.com/api3dao/signed-api/blob/e5e29dc21c7597967d5b87ba7849e38cbbfcacf6/packages/airnode-feed/package.json#L19

AFAIK, @Siegrift uses macOS, so that's why resulted images always creates signed-api target. I'm creating PR that fixing this.

Siegrift:
I see. Sorry @metobom I misunderstood your issue, because I saw that the error was saying it couldn't find the `signed-api.json` file. The fix from @bdrhn9 should fix the issue and I will release a patch.

PR #191, opened 2024-01-07
Create config directory in terraform templates

Siegrift:
Relates to https://github.com/api3dao/signed-api/issues/186#issuecomment-1874303494

We need to change the TF templates, because after the NPM publishing the `config` directory will no longer be part of the image.


bdrhn9:
> This was a simple changed so I just merged it.

Sorry I missed the review.

Issue #193, opened 2024-01-09
Make cache settings more generic


Siegrift:
Currently, the cache is convenient for the user, but very opinionated by setting `cdn-cache-control` which is not supported by AWS CloudFront. 

Also, the POST request returns the caching headers https://github.com/api3dao/signed-api/blob/0ce935772c082338263b76110cc9b95d0f9ae926/packages/signed-api/src/handlers.ts#L140 which it shouldn't.

PR #200, opened 2024-01-17
Support caching on AWS CloudFront

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/193

AWS CloudFront does not support caching based on custom header nor does recognize `cdn-cache-control`. The original issue mentioned making the cache headers more generic, but I don't think it's necessary. We can use both the standard `cache-control` for browser-cache (which is recognized by CloudFront) and the `cdn-cache-control`. 

The PR also removes these headers from POST response (which does not make sense).

### DIFF
--- a/packages/signed-api/README.md
+++ b/packages/signed-api/README.md
@@ -185,16 +185,17 @@ In case the endpoint should be publicly available, set the value to `null`.
 
 #### `cache` _(optional)_
 
-Configures the cache for the API endpoints.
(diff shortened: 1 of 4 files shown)


Siegrift:
@bdrhn9 LMK if you see some issues with this, but I think it's OK.

andreogle (approved):

Siegrift:
> Nit: Probably you are meaning CloudFront, so we can update issue title and description :)

Yeah, thanks :) Too many cloudXyz at this point for my brain.

PR #206, opened 2024-01-29
Remove pnpm from engine field

Siegrift:
I am a bit tired of the frequent updates of pnpm field, for example https://github.com/api3dao/signed-api/pull/205.

Morover, this can cause warning for consumers who use other package manager (e.g. npm or yarn) and it provides very little benefit. I suggest removing it altogether.


andreogle (approved):
🙌🏻

Issue #213, opened 2024-02-02
Experiment with Signed API without ELB


Siegrift:
It should be possible to deploy Signed API without an load balancer and expose its public IP. Note, that we never use the IP directly as we have the CDN put on top of it. Using the ELB seems extraneous because there is just a single instance it forwards the requests to.

This could decrease the AWS costs by a large margin as can be seen [on the invoice from January](https://github.com/api3dao/tasks/issues/518#issuecomment-1921086609).

The ELB returns 502 when the service is unavailable or under heavy load. I am suspicious whether it works correctly, because we see 502 errors but AWS reports low CPU and memory usage. But it serves as a natural SPAM protection because it will guard the service from potential spam attacks.

bbenligiray:
Why is this on hold?

Siegrift:
I added the label after I opened https://github.com/api3dao/signed-api/pull/218 which was the prototype, but with some heavy downsides. I added "on hold" label, because I wasn't sure whether we want to proceed with this or not.

Siegrift:
I've looked at the solution described in the link and I am not sure how it can be applied to our problem.

The problem discussed in the issue is to make the outbound traffic from the service come from a static IP address. For that one can use NAT gateway which translates the private IP address of the service with the static IP of the NAT gateway.

[One of the comments](https://stackoverflow.com/a/57977694) suggests whitelisting the whole subnet, but that seems to refer to the target service (called by the AWS service that needs to have the static IP). In fact, our service [allows traffic from any IP](https://github.com/api3dao/signed-api/blob/f7209ee459adfaf546129be7df9029989f4fb764/packages/signed-api/deployment/cloudformation-template.json#L161).

The problem we have is that service IP changes each time a task is redeployed, but we need it to be static to use it with CDN. To assign static IP, it seems we [need to use ELB](https://repost.aws/knowledge-center/ecs-fargate-static-elastic-ip-address).

bbenligiray:
Yeah I see, https://stackoverflow.com/a/73038635 only gives you an outbound static IP

Siegrift:
I discussed the problem with Stefi and we came up with an alternative, that I like the most. Have a single load balancer that distributes the traffic to different services based on path. This is currently done by CDN (but it's costly). 

[...]

PR #218, opened 2024-02-03
AWS Signed API deployment without load balancer

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/213

## Rationale

Unfortunately, It's [not possible to have a static IP for AWS Fargate](https://repost.aws/knowledge-center/ecs-fargate-static-elastic-ip-address). This means the IP changes after each restart/redeployment and each time we need to update the CDN routing. Also, the IP cannot be listed in CloudFormation output parameters and is only available in the currently deployed task configuration (ES2 -> Clusters -> Services -> Tasks). We can get the URL via AWS CLI though.

If we relax the requirement of static IP we can use the CloudFormation template in this PR. Each Signed API redeployment/restart would require immediate change in CDN distribution, incurring some delay while the update takes place.

During my research I haven't seen other "confirmed" solutions that work around the problem that the IP is bound to the task (not service). I saw threads like [this](https://stackoverflow.com/questions/68651457/how-to-route-traffic-to-ecs-fargate-instance-without-an-application-load-balance#comment121377541_68665985) or [this](https://serverfault.com/a/1037243) which support using the LB for this purpose.

### DIFF
--- a/packages/signed-api/config/signed-api.example.json
+++ b/packages/signed-api/config/signed-api.example.json
@@ -1,19 +1,12 @@
 {
   "endpoints": [
     {
-      "urlPath": "/real-time",
-      "authTokens": ["${REAL_TIME_ENDPOINT_AUTH_TOKEN}"],
-      "delaySeconds": 0
-    },
-    {
-      "urlPath": "/delayed",
+      "urlPath": "/real",
       "authTokens": null,
-      "delaySeconds": 15
+      "delaySeconds": 0
     }
   ],
-  "allowedAirnodes": [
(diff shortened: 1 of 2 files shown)


Siegrift:
We have a new CloudFormation template in https://github.com/api3dao/signed-api-deployments together with other CloudFormation files to create notification service, ALB and CloudFront, etc...

It is arguable whether we actually need to have the CloudFormation template in this repo. It might be useful when someone else wants to deploy the Signed API. That said, for that use case we should provide the minimal (unopinionated) deployment setup (e.g. without Loki forwarding).

Issue #207, opened 2024-01-31
Fatal error in Airnode feed


metobom:
Nodary Airnode feed process failed with [this](https://api3.grafana.net/explore?schemaVersion=1&panes=%7B%22spw%22:%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bairnode%3D%5C%220xc52EeA00154B4fF1EbbF8Ba39FDe37F1AC3B9Fd4%5C%22%7D%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%221706648610000%22,%22to%22:%221706654190000%22%7D%7D%7D&orgId=1) error and because we read config from raw Github URL and change the deployment file's location (from `candidate-deployments` to the `active-deployments`) after the deployment, CF tried to redeploy the app but wget kept throwing 404 error.

> CF tried to redeploy the app but wget kept throwing 404 error.

As a solution to this, I will update CF EntryPoint in a way to try `candidate-deployments` path first if fails try [active-deployments](https://raw.githubusercontent.com/api3dao/api-integrations/main/data/apis/nodary/deployments/active-deployments/api3-20240123-airnode-feed.json). 

@Siegrift for visibility.

Siegrift:
The error message doesn't tell much :/ 

I couldn't find related issues for `Check failed: is_clonable_js_type || is_clonable_wasm_type.`. Do we have access to the Nodary Airnode feed to check some metrics? Maybe it was some CPU/memory spike or memory leak.

In regards to Github URLs we need to [make sure that the production ones are immutable](https://api3workspace.slack.com/archives/C05S589E7B4/p1700910873307079?thread_ts=1700498242.003409&cid=C05S589E7B4).

metobom:
> Do we have access to the Nodary Airnode feed to check some metrics?

Unfortunately, we don't have any metrics.

bdrhn9:
>  Do we have access to the Nodary Airnode feed to check some metrics?

For debugging purpose, I activated `ECS CloudWatch Container Insights` to collect metrics from the containers. If we experience the issue again, it will be helpful. 

It's charged extra so generally it wouldn't be enabled. But it's easy to enable it in CF template:

```diff
"AppCluster": {
[...]

PR #223, opened 2024-02-08
Try using node-slim Docker image instead of alpine

Siegrift:
Relates to https://github.com/api3dao/signed-api/issues/207

One of the things we wanted to try was using a different Docker image. It seems unlikely this should cause the fatal error but is easy enough to try. Another possibility would be to migrate to Node@20, but that requires more testing (and possibly fixing some errors) so I am not doing it straight away, but I've opened a separate issue for that https://github.com/api3dao/signed-api/issues/224

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@
 
 # We use the alpine image because of its small size. The alternative considered was the "slim" image, but it is larger
 # and we already use alpine (without issues) in other projects, so the size reduction seems worth it.
-FROM node:18-alpine AS build
+FROM node:18-slim AS build
 WORKDIR /app
 RUN npm install -g pnpm
 # Copy just the "pnpm-lock.yaml" file and use "pnpm fetch" to download all dependencies just from the lockfile. This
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -39,13 +39,13 @@ LABEL application="airnode-feed" description="Airnode feed container"
...
-FROM node:18-alpine as airnode-feed
+FROM node:18-slim as airnode-feed
 WORKDIR /app
 ENV NODE_ENV=production
 
-RUN addgroup -S deployed-airnode-feed && \
-    adduser -h /app -s /bin/false -S -D -H -G deployed-airnode-feed deployed-airnode-feed && \
-    chown -R deployed-airnode-feed /app
+RUN addgroup --system deployed-airnode-feed && \
+    adduser --home /app --shell /bin/false --system --disabled-password --ingroup deployed-airnode-feed deployed-airnode-feed && \
```
andreogle, line 47: 👍🏻 for explicit flags

andreogle (approved):

Dockerfile
Siegrift: Yeah, previously I copied them from somewhere and didn't change them afterwards. After I changed the distro, the command error out that using `-s` is ambiguous so I changed these to explicit ones.

PR #233, opened 2024-02-13
Add wget dependency to airnode-feed image

### LINKED ISSUE #232: Add `wget` dependency to `airnode-feed` image for configuration file retrieval
`api3dao/api-integrations` generates a CloudFormation template that relies on `wget` [for fetching configuration files](https://github.com/api3dao/api-integrations/blob/c2d5c7d99c784f7b5fb78866cb9bf3555fa75159/frontend/src/Helpers/DownloadConfig.js#L239) for `airnode-feed`. Therefore, having the `wget` command available is crucial. However, the recently switched `node:18-slim` base image does not include this command by default, necessitating its installation. While installing this package at runtime is another option (with api-integrations generating a suitable script for runtime installation), it's preferable to include this installation step during image creation. This approach ensures that `wget` is readily available, clearly signaling its necessity for usage and preventing potential disruptions in the future , especially during base image changes.

bdrhn9:
Closes #232

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,11 @@ FROM node:18-slim as airnode-feed
 WORKDIR /app
 ENV NODE_ENV=production
 
+# Update package lists and install wget
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y wget && \
+    rm -rf /var/lib/apt/lists/*
+
 RUN addgroup --system deployed-airnode-feed && \
     adduser --home /app --shell /bin/false --system --disabled-password --ingroup deployed-airnode-feed deployed-airnode-feed && \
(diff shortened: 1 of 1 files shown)


andreogle (approved):

Siegrift (commented):
Thanks. I forgot that the custom entrypoint trick is not tested by the e2e test.

Siegrift:
That said, I am quite disappointed by all of the changes needed to update the Dockerfile to support the node-slim image.

andreogle:
I'm really surprised `wget` isn't included by default. I thought it was one of the base tools included in almost everything (I know `curl` isn't though)

PR #241, opened 2024-02-23
Log received signed data

### LINKED ISSUE #222: Log received data in INFO level
The pricing collector currently needs to call the API3 Signed APIs for data frequently and process the data in memory to create aggregated pricing value. This is a service that needs to be run and managed. An alternative solution would be to store the data in logs and have a simpler lambda that queries these logs and computes the value retrospectively.

bdrhn9:
Closes #222


packages/signed-api/src/handlers.ts
```diff
@@ -80,6 +82,14 @@ export const batchInsertData = async (
     return generateErrorResponse(400, message, detail ? { detail, signedData } : { signedData });
   }
 
+  if (env.LOG_APIDATA) {
```
Siegrift, line 85: I was thinking whether or not to put it after (or instead) [this line](https://github.com/api3dao/signed-api/blob/44232f8d2219dce9481ea5f022082be51e0e7321/packages/signed-api/src/server.ts#L21), but imo it's better to treat this separate so I agree with this being here.

packages/signed-api/src/handlers.ts
```diff
@@ -80,6 +82,14 @@ export const batchInsertData = async (
...
+    // Log only the required fields to use less space, do not log the signature for security reasons.
+    const sanitizedData = batchSignedData.map((data) =>
+      pick(data, ['airnode', 'encodedValue', 'templateId', 'timestamp'])
+    );
+    logger.info('Received signed data.', { data: sanitizedData });
```
Siegrift, line 90: nit: Maybe to emphasise that there was some logic done beforehand?

```suggestion
    logger.info('Received valid signed data.', { data: sanitizedData });
```

packages/signed-api/src/schema.ts
```diff
@@ -77,6 +77,7 @@ export const envBooleanSchema = z.union([z.literal('true'), z.literal('false')])
 export const envConfigSchema = z
   // Intentionally not using strictObject here because we want to allow other environment variables to be present.
   .object({
+    LOG_APIDATA: envBooleanSchema.default('false'),
```
Siegrift, line 80: The spelling feels a bit weird to me. I think it should be
```suggestion
    LOG_API_DATA: envBooleanSchema.default('false'),
```

Siegrift (commented):
Not sure If you wanted my review already, but I took a brief look. 

LGTM so far, but we also want to add `LOG_HEARTBEAT` to control the heartbeat logger, right?

bdrhn9:
> Not sure If you wanted my review already, but I took a brief look.

I'm planning to think about this PR on weekend however I couldn't find time, that's why I didn't add anyone for reviewing. It's appreciated, thank you 🙏🏼 . I applied your suggestions 💪🏼 


> but we also want to add LOG_HEARTBEAT to control the heartbeat logger, right?

Which would be better? Implementing it in this PR or creating new issue?

Siegrift (approved):
👍 LGTM

PR #245, opened 2024-02-28
Support skipping API calls

### LINKED ISSUE #238: Support `Skipping an API call` feature
OIS allows to skip API calls. When OIS is configured to do so, it triggers errors on the airnode-feed side due to the absence of implementation for this feature. For more details, refer to the code [here](https://github.com/api3dao/signed-api/blob/2eed67ecd69dfc0346c290768f3bdaacd22c6a30/packages/airnode-feed/src/api-requests/data-provider.ts#L15) and its airnode counterpart [here](https://github.com/api3dao/airnode/blob/59c8cd8c93f002460649e6c707abe943f0fb2b1d/packages/airnode-node/src/api/index.ts#L312).

aquarat:
This PR adds support for skipping API calls.

Closes #238


andreogle (approved):

packages/airnode-feed/src/api-requests/data-provider.ts
```diff
@@ -27,6 +27,10 @@ export const callApi = async (
       );
       logger.debug('Performing API call.', { processedEndpointParameters });
 
+      if (!endpoint.operation && isEmpty(endpoint.fixedOperationParameters)) {
+        return;
```
bdrhn9, line 31: Just to note, I'd prefer to mimic exactly same behaviour with airnode which directs output of pre-processing to post-processing. One can see [here](https://github.com/api3dao/airnode/blob/59c8cd8c93f002460649e6c707abe943f0fb2b1d/packages/airnode-node/src/api/index.ts#L315).
```suggestion
        return { data: processedEndpointParameters };
```

Nit: it'd be better to inform about skipping API call in debug mode like following: 
```ts
logger.debug('Skipping API call.', { processedEndpointParameters });
```

Siegrift: Agree with Bedirhan here. I'd also add to only print out:
```ts
logger.debug('Performing API call.', { processedEndpointParameters });
```

if we are actually making an API call.

aquarat: I've added logging statements for both skipped and executed API call paths, LMK if this is not what you wanted.

Thanks @bdrhn9 I don't know how I missed that :)

packages/airnode-feed/src/api-requests/data-provider.test.ts
```diff
@@ -104,4 +104,48 @@ describe(makeTemplateRequests.name, () => {
       url: 'https://api.nodary.io/feed/latestV2',
     });
   });
+
+  it('can skip an API call', async () => {
+    jest.clearAllMocks();
```
Siegrift, line 109: Is this required? We have [this](https://github.com/api3dao/signed-api/blob/73f53ef80cfcbf6f2418fa56194ab8833cac5da8/jest.config.js#L13) in the config which should do that automatically afaik.

packages/airnode-feed/src/api-requests/data-provider.ts
```diff
@@ -27,6 +27,13 @@ export const callApi = async (
       );
       logger.debug('Performing API call.', { processedEndpointParameters });
 
+      if (!endpoint.operation && isEmpty(endpoint.fixedOperationParameters)) {
+        logger.debug('Skipping API call.', { processedEndpointParameters });
+
+        return { data: processedEndpointParameters };
+      }
+
+      logger.debug('Performing API call.', { processedEndpointParameters });
```
Siegrift, line 36: You forgot to remove the call on line 28.

Siegrift (changes requested):
Really minor changes and it's good to go.

packages/airnode-feed/src/api-requests/data-provider.test.ts
aquarat: If I don't do this the test fails because `axios` has been called (it hasn't though, so it must be a leak from the previous test?). Tests should be independent/mocks should be reset, but for some reason that doesn't happen here.

packages/airnode-feed/src/api-requests/data-provider.ts
aquarat: 🙈, removed!

packages/airnode-feed/src/api-requests/data-provider.test.ts
aquarat: I think it's because `restoreMocks` in the config means jest will restore mocks to their initial implementation - but here we mock axios in the parent scope, so the initial implementation is that mock? and perhaps in those scenarios no reset occurs.

There is also a `resetMocks` config option.

Siegrift: I remember now. The problem is that mocks are not affecting calls to jest.mock here https://github.com/api3dao/signed-api/blob/238-support-skipping-api-call/packages/airnode-feed/src/api-requests/data-provider.test.ts#L16

If we use the newly implemented http module fro commons, we will not need that https://github.com/api3dao/signed-api/issues/247

Siegrift (approved):
👍 LGTM

packages/airnode-feed/src/api-requests/data-provider.test.ts
aquarat: Yes, so adding `resetMocks` to the jest config fixes this issue and doesn't break any other tests.

aquarat: Ah, I just saw your last comment, LMK if you're happy with enabling `resetMocks` @Siegrift

Siegrift (approved):
👍 LGTM, thanks for fixing the test.

packages/airnode-feed/src/api-requests/data-provider.test.ts
aquarat: Ah nice, thanks 👌

PR #246, opened 2024-02-28
Node 20 Bump

### LINKED ISSUE #224: Upgrade to Node@20
Relates to https://github.com/api3dao/signed-api/issues/207

This might help resolve the FATAL error reported by Node.js runtime. Also, per https://endoflife.date/nodejs we will need to do it anyway soon.

aquarat:
This bumps all Node 18 references to Node 20. 

I'm not sure if this needs to be tested beyond the CI tests?

Closes #224

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "signed-api",
   "version": "0.5.1",
   "engines": {
-    "node": ">=20.11.1"
+    "node": ">=18.19.1"
   },
   "scripts": {
     "build": "pnpm recursive run build",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "signed-api",
   "version": "0.5.1",
   "engines": {
-    "node": ">=18.19.1"
+    "node": ">=20.11.1"
   },
   "scripts": {
     "build": "pnpm recursive run build",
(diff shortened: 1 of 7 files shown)


andreogle (approved):

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "signed-api",
   "version": "0.5.1",
   "engines": {
-    "node": ">=18.19.1"
```
Siegrift, line 5: Let's leave the engines >=18.19.1. The problem is that these will cause warnings when you add the project as NPM dependency.

Siegrift (approved):
👍 LGTM, but I'd prefer if node version in package.json remains >=18.

> I'm not sure if this needs to be tested beyond the CI tests?

Nodary will try Airnode feed deployment and confirm nothing breaks (but I doubt anything would). 

We don't have to worry that much about Signed API because even if there is some breakage, we can easily revert to the previous Docker image and redeploy (we don't have such luxury with Airnode feeds).

package.json
aquarat: No problem, in all `package.json`s or just the root `project.json`?

Siegrift: All. When you install a package, only the one for the project is included.

Siegrift (approved):
👍 LGTM

Siegrift:
Merging because I want to release new Signed API version with this change included.

Issue #248, opened 2024-03-01
Ignore casing when comparing address


Siegrift:
Address can either be lowercase or EIP-55. Comparison should work with both (and their combinations).

PR #253, opened 2024-03-07
Ignore casing when comparing address

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/248

## Rationale

Make sure the Airnode address is checksummed before checking if it's valid. I think the nicest solution is to let it be checksummed by making a zod transformation. Initially, I wanted to lowercase this (in Auctioneer), but the problem is that addresses coming from web3 clients are checksummed (ethers, viem...) so it makes more sense to checksum them. This is something that should be included in https://github.com/api3dao/commons.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,7 +1,7 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const { join } = require('node:path');
 
-/*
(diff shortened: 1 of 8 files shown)


andreogle (approved):
👍🏻

PR #256, opened 2024-03-11
Use node least privilidged user, set NODE_ENV to production

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/242

TODO:
- [x] - Test AWS Airnode feed + Signed API deployment (because there may be issues with the entrypoint).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,12 +48,9 @@ RUN apt-get update && \
     apt-get install --no-install-recommends -y wget ca-certificates && \
     rm -rf /var/lib/apt/lists/*
 
-RUN addgroup --system deployed-airnode-feed && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -65,7 +62,11 @@ FROM build AS deployed-signed-api
 RUN pnpm --filter=@api3/signed-api --prod deploy deployed-signed-api
 FROM node:20-slim as signed-api
 WORKDIR /app
-ENV NODE_ENV=production
+
+# Update package lists and install wget
+RUN apt-get update && \
```
Siegrift, line 67: When testing the package I realized this is missing in the Signed API image. So 0.6.0 Signed API is broken, but it's not yet utilized.

Issue #209, opened 2024-01-31
Move configuration to separate README


Siegrift:
I suggested this [a while ago](https://github.com/api3dao/signed-api/issues/109#issuecomment-1798844770) but ended up [not doing it](https://github.com/api3dao/signed-api/pull/125#issue-2003728953) to avoid the extra work.

Since then the README got more exhaustive with multiple sections being added. It makes more sense now to separate the configuration into separate file, for the same reasons mentioned [here](https://github.com/api3dao/oev-auctioneer/issues/33#issuecomment-1918632631). Do not forget tho change the links to spec. This change should be done for both Airnode feed and Signed API.

PR #257, opened 2024-03-11
Move configurations to separate markdown file

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/209

- [x] After this is finalized I need to change the spec links.


andreogle (approved):
We don't want to put these in the `config/` folders?

Siegrift:
That might actually be a better place. Good point.

Issue #259, opened 2024-03-11
Remove updateDelay parameter from Airnode feed


Siegrift:
Airnode feed supports [updateDelay](https://github.com/api3dao/signed-api/blob/74340efc3dcbf61da2dd91e4293289b91c195608/packages/airnode-feed/README.md?plain=1#L289) parameter. It's nice that the Airnode feed can send delayed data to Signed API, but this feature doesn't need to exist.

See: https://api3workspace.slack.com/archives/C05S589E7B4/p1710148925838459?thread_ts=1710147088.423849&cid=C05S589E7B4

PR #261, opened 2024-03-11
Remove updateDelay parameter from Airnode feed

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/259

## Rationale

This feature is implemented already at Signed API level. As of now, all Airnode feeds push real time data to Signed APIs (so use update delay 0) and there doesn't seem to be an use case for setting this to a non-zero value.


Siegrift:
cc: @metobom This is going to be a breaking change for api-integrations.

PR #268, opened 2024-03-21
Small code cleanup

Siegrift:
While working on https://github.com/api3dao/airseeker-v2/pull/213 I noticed a small cleanup chance.

### DIFF
--- a/packages/airnode-feed/src/index.ts
+++ b/packages/airnode-feed/src/index.ts
@@ -5,8 +5,6 @@ import { initializeState } from './state';
 import { loadConfig } from './validation/config';
 import { loadEnv } from './validation/env';
 
-const env = loadEnv();
-
 const setupUncaughtErrorHandler = () => {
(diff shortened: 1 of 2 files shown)


packages/airnode-feed/src/index.ts
```diff
@@ -29,6 +27,7 @@ const startAirnodeFeed = async () => {
   initializeState(config);
 
   void initiateSignedApiUpdateLoops();
+  const env = loadEnv();
```
andreogle, line 30: What's the significance of this change? I can see that it now only loads the `process.env` variables when the function is called (rather than at application start time), but does that impact anything?

andreogle (approved):

packages/airnode-feed/src/index.ts
Siegrift: No, it doesn't. I just prefer to load that as close to the usage as possible.

PR #275, opened 2024-03-26
Fix heartbeat 0.7.0

Siegrift:
I've created a v0.7 branch in order to backport this fix to 0.7.0... I unfortunately can't do it on the `main` branch, because there are other features already merged that I don't want to release.


Siegrift:
This PR was wrong - the tests were not passing. The correct one is https://github.com/api3dao/signed-api/pull/276

PR #280, opened 2024-03-27
Implement skip API call using processing v2

Siegrift:
Resolves https://github.com/api3dao/vitepress-docs/issues/715#issuecomment-2022100690


dcroote (approved):
Nice 👍

Issue #325, opened 2024-06-28
Allow Signed API to expose real-time data without signature


Siegrift:
This would enable writing real-time OEV bots as a backup for capturing the OEV.

Issue #327, opened 2024-06-28
Expose real-time data for alternative beacons


Siegrift:
Blocked by https://github.com/api3dao/signed-api/issues/326

Issue #306, opened 2024-05-15
Export Signed API schema


Siegrift:
See: https://github.com/api3dao/airseeker-v2/issues/308

We should define and export the Signed API schema for the consumers to use.

PR #330, opened 2024-07-01
Expose unsigned data

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/325

An example response from the API:
```
{"count":3,"data":{"0xebba8507d616ed80766292d200a3598fdba656d9938cecc392765d4a284a69a4":{"airnode":"0xbF3137b0a7574563a23a8fC8badC6537F98197CC","templateId":"0xcc35bd1800c06c12856a87311dd95bfcbb3add875844021d59a929d79f3c99bd","timestamp":"1719819852","encodedValue":"0x0000000000000000000000000000000000000000000000047432f4b29a180000"},"0x6f6acbdadaaf116c89faf0e8de1d0c7c2352b01cce7be0eb9deb126ceaefa6ba":{"airnode":"0xbF3137b0a7574563a23a8fC8badC6537F98197CC","templateId":"0x086130c54864b2129f8ac6d8d7ab819fa8181bbe676e35047b1bca4c31d51c66","timestamp":"1719819852","encodedValue":"0x000000000000000000000000000000000000000000000001944d352591748000"},"0x7944f22b40cc691a003e35db4810b41543a83781d94f706b5c0b6980e0a06ed7":{"airnode":"0xbF3137b0a7574563a23a8fC8badC6537F98197CC","templateId":"0x1d65c1f1e127a41cebd2339f823d0290322c63f3044380cbac105db8e522ebb9","timestamp":"1719819852","encodedValue":"0x00000000000000000000000000000000000000000000007e1a3b815ec22a0000"}}}
```

The response is identical to the full Signed data, except the signature is omitted, making the data unusable for base fee updates.

Note, that this makes it awkward for [typing the return response](https://github.com/api3dao/signed-api/issues/306), but the caller will know whether they query the full or unsigned data endpoint, so this should not be an issue. This feature will also be superceded by https://github.com/api3dao/signed-api/issues/327

### DIFF
--- a/packages/airnode-feed/config/secrets.example.env
+++ b/packages/airnode-feed/config/secrets.example.env
@@ -2,3 +2,5 @@
 WALLET_MNEMONIC=diamond result history offer forest diagram crop armed stumble orchard stage glance
 # Get it from core development team, otherwise you can use some publicly available data provider API instead of Nodary.
 NODARY_API_KEY=
+# The Signed API auth token allowing this Airnode feed to push the signed data.
+AIRNODE_FEED_AUTH_TOKEN=some-secret-token-for-airnode-feed
--- a/packages/signed-api/src/handlers.test.ts
+++ b/packages/signed-api/src/handlers.test.ts
@@ -1,5 +1,6 @@
 import { ethers } from 'ethers';
 import { omit } from 'lodash';
+import type Pool from 'workerpool/types/Pool';
 
 import { getMockedConfig } from '../test/fixtures';
 import { createSignedData, generateRandomBytes, generateRandomWallet } from '../test/utils';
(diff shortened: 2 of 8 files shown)


packages/airnode-feed/config/secrets.example.env
```diff
@@ -2,3 +2,5 @@
 WALLET_MNEMONIC=diamond result history offer forest diagram crop armed stumble orchard stage glance
 # Get it from core development team, otherwise you can use some publicly available data provider API instead of Nodary.
 NODARY_API_KEY=
+# The Signed API auth token allowing this Airnode feed to push the signed data.
+AIRNODE_FEED_AUTH_TOKEN=some-secret-token-for-airnode-feed
```
Siegrift, line 6: From the perspective of Airnode feed, this should be named `SIGNED_API_AUTH_TOKEN`, but then the reader needs to infer that these are the same, which I don't think is worth it.

packages/signed-api/src/handlers.test.ts
```diff
@@ -11,14 +12,17 @@ import { logger } from './logger';
...
 beforeEach(() => {
   jest.spyOn(configModule, 'getConfig').mockImplementation(getMockedConfig);
-  initializeVerifierPool();
+  workerPool = initializeVerifierPool();
 });
 
-afterEach(() => {
+afterEach(async () => {
   inMemoryCacheModule.setCache({});
+  await workerPool.terminate();
```
Siegrift, line 25: This allows the tests to stop gracefully.

bdrhn9 (approved):
Looks good 👍🏼

PR #332, opened 2024-07-01
Update pnpm

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/331

Just ran `pnpm i` with version@9


Siegrift:
Merging, because this was trivial change without altering the logic.

Issue #326, opened 2024-06-28
Make Airnode feed sign data for alternative templates


Siegrift:
For each template ID defined in the Airnode feed config, we should also sign the value for the alternative template ID created by `keccak256(abi.encodePacked(templateId))`.  This is required for better OEV recapture.

Siegrift:
We've talked on the call, that it would be beneficial to for Airnode feed to send the signature for alternate beacon as part of the current payload.

Issue #262, opened 2024-03-11
Do not push airnode address and beacon ID from Airnode feed 


Siegrift:
These feeds are redundant:
1. Airnode address - is specified in the URL, but because it's also sent in the payload we do [this check](https://github.com/api3dao/signed-api/blob/b2e668c5136c1b4268a7ae0c0b72cfe704b832f5/packages/signed-api/src/handlers.ts#L62).
2. Bacon ID - Can be derived from Airnode address and template ID.

PR #347, opened 2024-07-31
Implement Signed api v2

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/262
Closes https://github.com/api3dao/signed-api/issues/306
Closes https://github.com/api3dao/signed-api/issues/326
Closes https://github.com/api3dao/signed-api/issues/327

## Rationale

In preparation for new version of OEV Auctioneer, we need to serve beacon data for "oev" beacons, which are derived from the base feed beacons (the ones that Airseeker updates). These are derived [by hashing the template ID](https://github.com/api3dao/contracts/blob/c6311d60881380c75820dd70c2571f470edee522/contracts/api3-server-v1/Api3ServerV1OevExtension.sol#L152).

As discussed on call, this will be a breaking change for the Signed APIs, because it enforces a strict input payload. To roll this out without breakage the following needs to happen:
1. Signed API needs to be implemented to support both v1 and v2 (released from these new changes).
2. Signed APIs need to be redeployed
3. Airnode feeds will be redeployed

This is achieved by doing the following:
1. Airnode feed pushing schema has changed and Signed API tries to parse both v2 and v1 payload. It converts both to an internal representation that it persists in memory.
2. The endpoint has a new flag `isOev` which specifies whether the endpoint returns OEV data (for the derived beacons) or the base ones (for Airseeker usage). This field is required for explicitness, as it's a crucial setting.


bdrhn9 (approved):
LGTM 🔥

Siegrift:
Added the OEV team to take a look as well.

packages/signed-api/config/signed-api.example.json
andreogle, line 1: Is it worth having an OEV endpoint here?

packages/signed-api/src/transform-payload.ts
```diff
@@ -0,0 +1,71 @@
...
+      [airnode]: { ...cache.airnodeToTemplateIdToBeaconId[airnode], [templateId]: beaconId },
+    },
+  });
+  return beaconId;
+};
+
+export const transformAirnodeFeedPayload = (
+  payload: SignedApiBatchPayloadV1 | SignedApiBatchPayloadV2
+): InternalSignedData[] => {
+  if ('airnode' in payload) {
```
andreogle, line 41: Some comment here might be useful to explain the alternative code path and why it's checked like this

andreogle (approved):

packages/e2e/test/signed-api.feature.ts
```diff
@@ -74,3 +76,27 @@ test('handles both EIP-55 and lowercased addresses', async () => {
 
   expect(eip55AirnodeResponse.data).toStrictEqual(lowercasedAirnodeResponse.data);
 });
+
+test.only('returns OEV beacons from OEV endpoint', async () => {
```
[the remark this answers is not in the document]
Siegrift: I wonder why ESLint doesn't complain about this.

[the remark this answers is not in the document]
Siegrift: Figured it out. See: https://github.com/api3dao/eslint-plugin-commons/issues/38

packages/signed-api/config/signed-api.example.json
Siegrift: Added.

Issue #364, opened 2024-09-09
Log full incoming signed data with delay


Siegrift:
Currently, we only log the signed data without signature to avoid people who have access t logs to mis-use the data that should otherwise be private (or reserved for OEV usage).

The data is used for internal purposes, such as pricing or other related statistics. What we can do is log the data along with the  signature with 5 minute delay (at which point the data is already exposed through public endpoint of Signed API).

See also: https://api3workspace.slack.com/archives/C05TQPT7PNJ/p1725547913985369

PR #365, opened 2024-09-10
Log full data with delay

Siegrift:
Closes https://github.com/api3dao/signed-api/issues/364

### DIFF
--- a/packages/signed-api/config/configuration.md
+++ b/packages/signed-api/config/configuration.md
@@ -52,7 +52,7 @@ Default: `json`.
 
 Enables or disables logging of the API data at the `info` level. When set to `true`, received valid signed data will be
 logged with the fields `airnode`, `encodedValue`, `templateId`, `timestamp` and `signature`. The logging of this data is
-delayed to make sure people with access to the logs won't be able to misuse the beacon data. Options:
+delayed by 5 minutes to make sure people with access to the logs won't be able to misuse the beacon data. Options:
 
 - `true` - Enables logging of the API data.
 - `false` - Disables logging of the API data.
--- a/packages/signed-api/config/configuration.md
+++ b/packages/signed-api/config/configuration.md
@@ -51,8 +51,8 @@ Default: `json`.
 ### `LOG_API_DATA` _(optional)_
 
 Enables or disables logging of the API data at the `info` level. When set to `true`, received valid signed data will be
-logged with the fields `airnode`, `encodedValue`, `templateId`, and `timestamp`. The `signature` field is intentionally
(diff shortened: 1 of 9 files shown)


packages/signed-api/src/handlers.ts
```diff
@@ -7,19 +7,21 @@ import { loadEnv } from './env';
...
   GetListAirnodesResponseSchema,
   GetSignedDataResponseSchema,
   GetUnsignedDataResponseSchema,
+  InternalSignedData,
   PostSignedDataResponseSchema,
 } from './types';
 import { extractBearerToken, generateErrorResponse, isBatchUnique } from './utils';
 
-const env = loadEnv();
+const LOG_API_DATA_DELAY_MS = 5 * 60 * 1000;
```
Siegrift, line 23: I think a hardcoded delay is good enough.

bdrhn9: Yes it looks okay unless we set [delaySeconds](https://github.com/api3dao/signed-api-deployments/blob/813303ae23c588d5b1a3894af6112fd679e0a54b/deployments/ankr/signed-api.json#L25) to more than 5 minutes.

packages/signed-api/config/configuration.md
```diff
@@ -51,8 +51,8 @@ Default: `json`.
 ### `LOG_API_DATA` _(optional)_
 
 Enables or disables logging of the API data at the `info` level. When set to `true`, received valid signed data will be
-logged with the fields `airnode`, `encodedValue`, `templateId`, and `timestamp`. The `signature` field is intentionally
-excluded for security reasons. Options:
+logged with the fields `airnode`, `encodedValue`, `templateId`, `timestamp` and `signature`. The logging of this data is
+delayed to make sure people with access to the logs won't be able to misuse the beacon data. Options:
```
bdrhn9, line 55: Nit: maybe we can mention how much logging is delayed because it is already hard-coded to 5 minutes.

bdrhn9 (approved):
Looks good to me 👍🏼

Siegrift:
> Yes it looks okay unless we set [delaySeconds](https://github.com/api3dao/signed-api-deployments/blob/813303ae23c588d5b1a3894af6112fd679e0a54b/deployments/ankr/signed-api.json#L25) to more than 5 minutes.

Yeah, but such data is practically unusable so I don't see us doing that.

packages/signed-api/config/configuration.md
Siegrift: Sure.

PR #420, opened 2025-02-17
Add `/status` endpoint and specify certified airnode addresses

### LINKED ISSUE #397: Add `/healtcheck` endpoint
As we are planning to make partner API providers deploy Signed API, we need to have a control mechanism over those deployments. For that, we need a new endpoint named `/healtcheck` that returns the contents of the Airnode feed heartbeat.

**The interface for the heartbeat without signature:**

```
export interface UnsignedHeartbeatPayload {
  airnode (of the allowedAirnode with `isCertified` true): string;
  stage: string;
  version (of Signed API): string;
  currentTimestamp: string;
  deploymentTimestamp: string;
  configHash (of unpopulated config): string;
}
```

Note that https://github.com/api3dao/signed-api/issues/406 must be completed before this one due to `airnode` field in the `UnsignedHeartbeatPayload`.

**The interface for the HeartbeatPayload:** 
```
export interface SignedHeartbeatPayload {
  airnode (of the allowedAirnode with `isCertified` true): string;
  stage: string;
  version (of Signed API): string;
  currentTimestamp: string;
  deploymentTimestamp: string;
  configHash (of unpopulated config): string;
  signature (where message is UnsignedHeartbeatPayload): string;
}
```

**As a result**, the `/healtcheck` endpoint would return the following JSON:
```
{
  "airnode": "0x...",
  "stage": "aws",
  "version": "2.1.0",
  "currentTimestamp": "1736238429",
  "deploymentTimestamp": "1736238425",
  "configHash": "0x...",
  "signature": "0x..."
}
```

### LINKED ISSUE #406: Specify owner for Signed API
We are planning to make partner APIs deploy Signed API. Their Signed API URL will look like the following:

```
https://signed-api.<API_NAME>.com
```
Signed API will be under their domain but also their Airnode address should be verifiable as well.

* Add ~owner~ `isCertified` flag to [allowedAirnodes](https://github.com/api3dao/signed-api/blob/3b44707e77c6128f805073505fdaa5917cead0f8/packages/signed-api/config/signed-api.example.json#L29). Only one Airnode address can be the owner.

```
{ "address": "0xbF3137b0a7574563a23a8fC8badC6537F98197CC", "~owner~" "isCertified": true, "authTokens": ["${AIRNODE_FEED_AUTH_TOKEN}"] }
```
* List allowed Airnodes and show the owner in `/` endpoint.

So it would look like the following:
```
{
  "allowedAirnodes": [
    ...
  ],
  "certifiedAddress": "0xf52d..."
}
```

It is open to discussion, lmk.

bdrhn9:
Closes #397 
Closes #406 

Unlike the https://github.com/api3dao/signed-api/issues/406, I listed only `certifiedAirnodeAddresses` (airnodes with `isCertified: true`) since we don’t need signed-api config details. The goal is to verify first-party Airnodes, and the market only requires certified addresses.

### DIFF
--- a/packages/signed-api/config/configuration.md
+++ b/packages/signed-api/config/configuration.md
@@ -219,14 +219,14 @@ or
 
 ```jsonc
 // Allows pushing signed data only for the specific Airnode. No authorization is required to push the data.
-"allowedAirnodes": [ { "address": "0xB47E3D8734780430ee6EfeF3c5407090601Dcd15", "authTokens": null } ]
+"allowedAirnodes": [ { "address": "0xB47E3D8734780430ee6EfeF3c5407090601Dcd15", "authTokens": null, "isCertified": true } ]
 ```
 
 or
 
 ```jsonc
 // Allows pushing signed data only for the specific Airnode. The pusher needs to authorize with one of the specific tokens.
-"allowedAirnodes": { "address": "0xbF3137b0a7574563a23a8fC8badC6537F98197CC", "authTokens": ["some-secret-token-for-airnode-feed"] }
+"allowedAirnodes": { "address": "0xbF3137b0a7574563a23a8fC8badC6537F98197CC", "authTokens": ["some-secret-token-for-airnode-feed"], "isCertified": true }
 ```
 
 #### `allowedAirnodes[n]`
@@ -244,6 +244,12 @@ The nonempty list of
 
 To allow pushing data without any authorization, set the value to `null`. The API validates the data, but this is not
 recommended.
+
+##### `isCertified`
+
+A boolean flag indicating whether the Airnode is first-party verified. When set to true, the Airnode address will be
+included in the `/status` endpoint response under `certifiedAirnodeAddresses`. This proves that the deployer of the
+Signed API instance owns/controls these Airnodes, enabling clients to verify first-party data sources.
 
 #### `stage`
 
--- a/packages/signed-api/src/handlers.ts
+++ b/packages/signed-api/src/handlers.ts
@@ -1,4 +1,5 @@
 import { signedApiBatchPayloadV1Schema, signedApiBatchPayloadV2Schema } from '@api3/airnode-feed';
+import { createSha256Hash, serializePlainObject } from '@api3/commons';
 import { go, goSync } from '@api3/promise-utils';
 import { isEmpty, omit, pick } from 'lodash';
 
@@ -17,10 +18,14 @@ import type {
   GetUnsignedDataResponseSchema,
   InternalSignedData,
   PostSignedDataResponseSchema,
+  GetStatusResponseSchema,
 } from './types';
 import { extractBearerToken, generateErrorResponse, isBatchUnique } from './utils';
 
 const LOG_API_DATA_DELAY_MS = 5 * 60 * 1000;
+
+// Initialize deployment timestamp when the application starts
+const DEPLOYMENT_TIMESTAMP = Math.floor(Date.now() / 1000).toString();
 
 // Accepts a batch of signed data that is first validated for consistency and data integrity errors. If there is any
 // issue during this step, the whole batch is rejected.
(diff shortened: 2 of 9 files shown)


packages/signed-api/src/types.ts
```diff
@@ -21,6 +21,15 @@ export type PostSignedDataResponseSchema = {
 
 export type GetListAirnodesResponseSchema = { count: number; 'available-airnodes': string[] };
 
+export type GetStatusResponseSchema = {
+  stage: string;
+  version: string;
+  currentTimestamp: string;
+  deploymentTimestamp: string;
+  configHash: string;
+  certifiedAirnodeAddresses: string[];
```
bdrhn9, line 30: `certifiedAirnodeAddresses` is defining variable better however I think it is more convenient to use `certifiedAirnode` in the repository. Here other usages:
https://github.com/api3dao/signed-api/blob/d91f6c3ecd6feaf5db578b993fc731dc2e41b2c2/packages/signed-api/src/types.ts#L22
https://github.com/api3dao/signed-api/blob/d91f6c3ecd6feaf5db578b993fc731dc2e41b2c2/packages/signed-api/src/types.ts#L12

Wdyt?

packages/signed-api/config/configuration.md
```diff
@@ -245,6 +245,12 @@ The nonempty list of
 To allow pushing data without any authorization, set the value to `null`. The API validates the data, but this is not
 recommended.
 
+##### `isCertified`
+
+A boolean flag indicating whether the Airnode is first-party verified. When set to true, the Airnode address will be
+included in the `/status` endpoint response under `certifiedAirnodeAddresses`. This proves that the deployer of the
```
Siegrift, line 251: I have a bit of trouble with the wording. Technically, the Signed API deployer can set this to whatever value they want so they could make it look like they control also Airnodes they don't in reality. It doesn't make sense to do so in practice, but the docs should mention this is possible.

packages/signed-api/src/handlers.ts
```diff
@@ -17,11 +18,15 @@ import type {
...
   InternalSignedData,
   PostSignedDataResponseSchema,
+  GetStatusResponseSchema,
 } from './types';
 import { extractBearerToken, generateErrorResponse, isBatchUnique } from './utils';
 
 const LOG_API_DATA_DELAY_MS = 5 * 60 * 1000;
 
+// Initialize deployment timestamp when the application starts
+const DEPLOYMENT_TIMESTAMP = Math.floor(Date.now() / 1000).toString();
```
Siegrift, line 28: Use [this](https://github.com/api3dao/signed-api/blob/b054c505b3fe3bafcc46ff4064b6dcd7356066cb/packages/airnode-feed/src/state.ts#L10) instead.

packages/signed-api/src/types.ts
Siegrift: I guess you meant `certifiedAirnodes` in plural (as it's an array). 

I see the variables using "Airnode address" and "Airnode" are used inconsistently. My take on this is that "Airnode address" is quite long so and it's quite clear what "Airnode" refers to (Burak also uses term "Airnode" in contracts [majority of the times](https://github.com/search?q=repo%3Aapi3dao%2Fcontracts%20airnodeAddress&type=code)). 

I think this is a very minor issue and I would not even bother refactoring this. I'd use the name `certifiedAirnodes`.

Siegrift (changes requested):
Left a few small comments

packages/signed-api/src/handlers.ts
bdrhn9: This is in `airnode-feed` application, so we cannot access state of the airnode-feed right?

Siegrift: 🤦 Right, I remembered we had stored this already so I just did a quick search in code for it. Disregard my comment.

Siegrift (approved):
👍 LGTM

PR #417, opened 2025-02-14
Bump airnode to v0.15.0

dcroote:
A downstream fix for https://github.com/api3dao/data-feed-reader-example/issues/199 since `airnode-feed` is pulling in `airnode-node` v0.14.2 which is pulling in commons ^0.5.0

image

PR #421, opened 2025-02-17
Add ts everywhere

Siegrift:
When I ran `pnpm run create-release:npm minor` the script failed on TSC checks because `tsc` was unavailable in some subpackages. I'm adding it expllicitely in this PR so that I can proceed with https://github.com/api3dao/signed-api/pull/417#issuecomment-2662221767

### DIFF
--- a/packages/airnode-feed/package.json
+++ b/packages/airnode-feed/package.json
@@ -46,6 +46,7 @@
   "devDependencies": {
     "@types/express": "^5.0.0",
     "@types/lodash": "^4.17.15",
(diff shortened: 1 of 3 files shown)


Siegrift:
Also, I guess we'll do the release after https://github.com/api3dao/signed-api/pull/420

Siegrift:
> After pnpm run build on the root of the project, the current main branch worked well to me. Can you try?

Mhh, strange... It works for me now as well. I also tried pruning all of the node_modules and did a clean install and the command worked... I am not sure why I got the error before...

PR #423, opened 2025-02-17
Move /status to the root / endpoint and list available Airnodes under /airnodes

### LINKED ISSUE #422: Move `/status` to the root `/` endpoint and list available Airnodes under `/airnodes`
Currently, the `/` endpoint returns a list of available Airnodes. However, when someone visits `signed-api.<domain>`, we want them to see certified airnode addresses without needing to specify any additional path. Therefore, I propose moving the `/status` endpoint to `/` and relocating the current `/` endpoint under `/airnodes`.

As part of these changes, it would also be better to switch the `available-airnodes` key from `kebab-case` to `camelCase` [here](https://github.com/api3dao/signed-api/blob/3f83dc03df0deb07312a2a7520531b177fd9f94c/packages/signed-api/src/types.ts#L22). Additionally, using `data` as the key name might be more appropriate, as the route itself `/airnodes` already clearly defines its purpose.

bdrhn9:
Closes #422

### DIFF
--- a/packages/performance-test/airnode-feed/create-config.ts
+++ b/packages/performance-test/airnode-feed/create-config.ts
@@ -90,8 +90,8 @@ async function main() {
   configTemplate.signedApis[0]!.url = targetSignedApiUrl;
   configTemplate.ois[0]!.apiSpecifications.servers[0]!.url = sourceSignedApiUrl;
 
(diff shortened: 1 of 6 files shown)


Siegrift (approved):
👍 LGTM

bdrhn9:
I don't like force-pushing but I had to do it for this PR.

PR #451, opened 2025-04-04
fix: use pnpm version from package.json

dcroote:
Also the v2 action has [stopped working](https://github.com/pnpm/action-setup/tree/v4/?tab=readme-ov-file#warning-upgrade-from-v2)


Siegrift (approved):
👍 LGTM

PR #457, opened 2025-04-27
Skip operation effect validation for endpoints using `Skip API call`

### LINKED ISSUE #456: Skip operation effect validation for endpoints using `Skip API Call` feature
The [current](https://github.com/api3dao/signed-api/blob/3b428df082871ad938c0d9dc9cf3331ec9ea60cd/packages/airnode-feed/src/validation/schema.ts#L161) trigger reference validation enforces that all templates in a batch must produce identical operation effects. However, this validation becomes problematic for endpoints utilizing the [Skip API Call](https://github.com/api3dao/signed-api/issues/238) feature.

When `Skip API Call` is enabled, we're explicitly bypassing the standard API call flow. This typically occurs when the OIS specification cannot adequately describe the intended API call, when generating data without external API calls (such as random numbers), or when implementing custom processing flows.

In these scenarios, pre/post-processing scripts will inherently produce variable results - by nature. The scripts might make independent API calls, generate dynamic values, or implement specialized transformations. The current operation effect validation incorrectly blocks these legitimate use cases.

The proposed solution skips operation effect validation for endpoints using `Skip API Call` feature.

bdrhn9:
Closes #456 

For visibility: @metobom


packages/airnode-feed/src/validation/schema.ts
```diff
@@ -167,33 +167,53 @@ const validateTriggerReferences: SuperRefinement<{
...
-  for (const signedApiUpdate of triggers.signedApiUpdates) {
+  for (const [signedApiUpdateIndex, signedApiUpdate] of triggers.signedApiUpdates.entries()) {
     const { templateIds } = signedApiUpdate;
 
+    // Verify all template IDs actually exist in the templates object
+    const referenceErrors = templateIds.map((templateId, templateIdIndex) => {
+      if (templates[templateId] === undefined) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: `Template ID "${templateId}" is not defined in the config.templates object`,
```
Siegrift, line 178: nit:
```suggestion
          message: `Template ID "${templateId}" is not defined`,
```

if we ever refactor the config it can go stale.

packages/airnode-feed/src/validation/schema.ts
```diff
@@ -167,33 +167,53 @@ const validateTriggerReferences: SuperRefinement<{
 }> = async (config, ctx) => {
   const { ois: oises, templates, endpoints, triggers } = config;
 
-  for (const signedApiUpdate of triggers.signedApiUpdates) {
+  for (const [signedApiUpdateIndex, signedApiUpdate] of triggers.signedApiUpdates.entries()) {
     const { templateIds } = signedApiUpdate;
 
+    // Verify all template IDs actually exist in the templates object
+    const referenceErrors = templateIds.map((templateId, templateIdIndex) => {
+      if (templates[templateId] === undefined) {
```
Siegrift, line 175: nit: 

```suggestion
      if (!templates[templateId]) {
```

packages/airnode-feed/src/validation/schema.ts
bdrhn9: > if we ever refactor the config it can go stale.

You're right but current version clearly points where the issue arises. I'd recommend to keep it.

